PR #26, opened 2016-08-29
Initial support for OS X

bbustin:
I created an installer and updated the Graphtec.py file to work on Mac OS X. It is hopefully done in a way that does not break things on other platforms.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,20 +63,11 @@ Windows (untested):
 
 
 Mac OS X (untested)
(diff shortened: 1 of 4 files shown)


jnweiger:
Cool!
Thanks for supporting mac. I have several users waiting for this here at FabLab Nürnberg!

bbustin:
Greetings from Oakland, California. Glad I could help. :-)

PR #41, opened 2017-04-25
Adding support for cameo 3

goopypanther:
Added USB PID/VIDs and margins for cameo 3 machine, tested and working.
This is just basic support, things like automatic autoblade adjustment and the second toolholder are still in the works.

Also added a new makefile target (install-local) that puts the scripts in the inkscape extensions folder in your home directory if you only want to install for current user (I find this easier for development).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,9 @@ VERS=$$(echo '<xml height="0"/>' | python ./sendto_silhouette.py --version /dev/
 DEST=$(DESTDIR)$(PREFIX)/share/inkscape/extensions
 UDEV=$(DESTDIR)/lib/udev
 
+# User-specifc inkscape extensions folder for local install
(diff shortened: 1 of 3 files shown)


EtherGraf:
Cannot test it, because I have no cameo3 around. But looks as nothing got broken. 
The install-local build target is a nice idea. 

I will merge this to master trunk.

lgrawet:
Hi @JeremyRuhland 

Have you written any piece of code for autoblade yet ?

Thanks

goopypanther:
No, I have not touched this in quite a while. I've uncovered some bugs too, placing lines in certain areas results in the cutter moving the entire pattern in unexpected directions across the work surface and frequently the rightmost edge of a pattern is never cut.
If I get a chance I'll try investigating exactly what the behavior is.

PR #45, opened 2017-06-23
multiple actions (e.g. score then cut)

lexelby:
This  adds a new extension, "Silhouette Multiple Actions".  Much like in Silhouette Studio, you can assign different actions (cut, score, pen) to each color of object in your drawing.  Of course, inkscape-silhouette is way better than Silhouette Studio. :)

How is this useful?  Consider [this SVG](https://cdn.rawgit.com/lexelby/fablab-polyhedra/master/panels/triangles_lex.svg):

triangle panels with tabs and slots

In the Action 1 tab, I set the color to red, the action to Cut, and the pressure to 2.  This lightly scores the lines, making them easy to crease precisely.

In the Action 2 tab, I set the color to blue, the action to Cut, and the pressure and speed to 0.  This cuts the shapes out.

The result is a nicely scored/cut page all in one go.  Without this extension, I'd have to select all the red paths, run "send to silhouette", set it to score, select blue, etc.  The result is a bunch of triangles with slots and tabs that you can put together into 3d shapes!

By setting the "pause" to a few seconds, you get enough time between actions to pause the machine and swap out tools.  This way, you could change knife depth or switch to a pen.

How it works: this extension gathers up the parameters in each tab separately.  It searches the document for objects with a matching stroke color, then invokes the "send to silhouette" extension directly once per action.  Kind of a hack, but it gets the job done. :)  In theory this functionality could simply be merged into "send to silhouette", but it's easier to understand how this works when it's in a separate file.

Note that I had to fix an existing bug while I was at it.  If you selected individual paths in a group and used "send to silhouette", it didn't properly apply the group's transforms to the paths.  This is because the groups weren't processed so their transforms weren't composed by the recursive algorithm.  I borrowed [a technique from inkscape-embroidery](https://github.com/lexelby/inkscape-embroidery/blob/master/embroider.py#L197).


lexelby:
cc @EtherGraf 

Your overcut feature is awesome!  It's been the difference between success and failure on more than one of my projects.  I thought you might find this pull request interesting.

EtherGraf:
Have not yet found time to test your patches.

I like your idea of creating a second extension. This separates your extension from the cutting part. But on the other hand this means a near full duplication of the extension gui. Future modifications needs to be done on both extensions. My ongoing branch minimizetraveling will cause already the first conflict.

The bugfix looks good. Lets test it.

lexelby:
Yup, agreed.  It kind of pained me to write it, especially the part where I had to copy-paste the whole tab 5 times.

Here's what I really want to do: write a GUI in Python.  It would pull all the stroke colors present in the document and present them to the user as a drop-down.  It would parse `sendto_silhouette.inx` and use that to construct a dialog.  That way, there would be no duplication of the UI and it would always be up-to-date with new parameters.

I did something somewhat similar with [inkscape-embroidery](https://github.com/lexelby/inkscape-embroidery)'s params dialog.  It reads the existing stitching parameters attached to each selected path and populates a wxGTK dialog.  It all works pretty nicely and steps around the issue where we can't dynamically populate the extension settings dialog that inkscape itself displays.

Anyway design is right there in my head, and it'd be awesome, but I just don't have the time to dedicate to writing it right now.  So I did this semi-hack just to get something working so that I could cut some shapes out for my children. :)

lexelby:
As to the bugfix, it seems to work for me, although I'm not quite sure how to test the viewbox thing.  I _think_ it's correct to multiply the docTransform on the left, but I'm not quite sure.

I could pull the fix out into its own PR and base this one on that PR, if you'd prefer?

lexelby:
Another thing a dedicated UI would make possible would be saving/loading presets.

lexelby:
Still a WIP, but I found some free time and rewrote this as a standalone GUI!

When started, this extension reads through the SVG and produces a list of stroke colors.  It presents a window with those stroke colors and lets you choose what to do with each color.  When you select a row, the settings pane is populated and allows you to fill in all of the normal parameters from "send to silhouette".

You can repeat this for each color, and the parameters are saved separately for each.  You can re-order the colors by selecting one and clicking the "move up" or "move down" buttons.  You can tell it to ignore a color entirely by unchecking its "perform action?" box.

Finally, you can save everything (including colors, order, checkboxes, and settings) as a preset by entering a preset name in the combo box and clicking "add".  You can also load a preset by choosing one from the drop-down and clicking "load".

To perform the actions, click "execute".  This also saves a hidden preset called "\_\_LAST\_\_", which is automatically loaded next time you run the extension.

To do:
* Test.  I haven't actually run this on my machine yet.
* When loading presets, it trusts the colors from the preset even if they don't exist in the current document.  It should filter out colors from the preset that aren't present in the SVG and add any missing colors.

lexelby:
Tested, found and fixed a few bugs, and then fixed up the issue with loading a preset with different colors.  This is ready to merge!

Future work that might be nice for a later PR: run `sendto_silhouette.py` in a background thread so that the wxgtk app can continue functioning properly.  Display a dialog showing which step we're on, with a cancel button that kills `sendto_silhouette.py`.  The cutter will still have to finish out its buffer before it cancels, but this can still save a lot of time if you know a cut job has gone awry.

myselfhimself:
Hello, is someone going to merge that some time?
The feature seems just awesome.. I want to prepare folds and outer cuts for my chocolate biscuits packaging... :-/

myselfhimself:
Hello,
I think that the xmltodict python package should be added somewhere in the setup.py, requirements.txt ... 
When trying to run the extension by clicking in Inkscape Extensions menu on my ubuntu 16.04 install, after clicking Apply, I get the following error in a dialog box:
"Traceback (most recent call last):
  File "silhouette_multi.py", line 12, in <module>
    import xmltodict
ImportError: No module named xmltodict"

myselfhimself:
This requires "sudo apt-get install python-wxgtk3.0 python-xmltodict"

myselfhimself:
After installing the aforementioned dependencies, I cannot go past this step (or I do not think any cool UI, that the PR adds), see attached image:
only_possible_step

myselfhimself:
I had the wrong branch sorry, now I have downloaded, run the setup.py and put the *.py *.inx and silhouette/ directory into /usr/share/inkscape/extensions (deleting the former silhouette extension files beforehand). I have opened the above triangles_lex.svg file and been through the steps.

Now, after the last comment's dialog, I get the following error in a dialog:

`Traceback (most recent call last):
  File "silhouette_multi.py", line 781, in <module>
    e.affect()
  File "/usr/share/inkscape/extensions/inkex.py", line 268, in affect
    self.effect()
  File "silhouette_multi.py", line 668, in effect
    self.frame = SilhouetteMultiFrame(colors=self.objects_by_color.keys(), run_callback=self.run, options=self.options)
  File "silhouette_multi.py", line 269, in __init__
    self.notebook = ParamsNotebook(self, wx.ID_ANY)
  File "silhouette_multi.py", line 101, in __init__
    self.load_inx()
  File "silhouette_multi.py", line 107, in load_inx
    self.inx = xmltodict.parse(inx_file, force_list=('param',))
  File "/usr/lib/python2.7/dist-packages/xmltodict.py", line 225, in parse
    **kwargs)
TypeError: __init__() got an unexpected keyword argument 'force_list'
`

myselfhimself:
Ok, my Ubuntu xmltodict package is somewhat too old. Here is [a related ticket on the xmltodict project](https://github.com/martinblech/xmltodict/issues/123#issuecomment-168707831).
I managed to install a more recent package with `sudo -H python2 -m pip install xmltodict `, now I can see the GUI. Yeeha! It would be nice to impose a minimum version, or check for an xmltodict minimum version at runtime and warn if needed.

myselfhimself:
This additional extension works well now, with added packages. **Thank you!**
It seems that the trimming on the margin on both of my 2 jobs does only work on one of them.. I have to try again and confirm..

multi_action_screen

After some time using, I feel that it would be more practical if the first setup dialog regarding command output without running them, should be a console or checkbox somewhere on the main dialog (eg. below the Execute button), or after the execute button was clicked.

lexelby:
Hi @myselfhimself, thanks for testing my branch!  Sorry for my radio silence, but I'm glad you figured out the troubles you were having.  I agree that a proper `requirements.txt` would be good, but I haven't gotten around to adding one.  Thanks for the note about the version of `xmltodict`.

I see what you mean about the "trim margins" checkbox.  The problem you're seeing is an artifact of the way `silhouette_multi` works.  In essence, it pretends like you've selected all objects of the first color and run the normal "send to silhouette" extension, then selected all objects of the second color and run "send to silhouette" again, etc.  It does several runs of "send to silhouette" all in a row for you.

This means that "trim margins" will effectively trim the margins on each color _individually_!  You'll get a different amount of trimming per color which is almost certainly not what you want.  I'd recommend not using that option with my "multiple actions" extension.  If you really want the margins trimmed off, move the objects on the canvas.

You're probably wondering, why does "multiple actions" present the trim option at all if it's a bad idea?  Because "multiple actions" loads the option list from the extension definition file (`sendto_silhouette.inx`) and generates the settings GUI automatically.  In general it's a good approach, but obviously there's a pitfall :)

> After some time using, I feel that it would be more practical if the first setup dialog regarding command output without running them, should be a console or checkbox somewhere on the main dialog (eg. below the Execute button), or after the execute button was clicked.

I see your point.  That option is really only there for debugging the "multiple actions" extension itself.  You're not likely to need it in day-to-day usage.  I actually never see that little initial dialog because I've bound ctrl-shift-s to "Silhouette Multiple Actions (No Preferences)" in Inkscape's preferences under Keyboard Shortcuts.  Hitting ctrl-shft-s brings me directly to the multiple actions extension's GUI window.

jnweiger (approved):
I like the idea of a separate tool with a separate GUI. Grat work!

Although this opens the portability question. What happens on windows and macos machines? Are there proper error messages that guide users to the needed dependencies?

EtherGraf:
Just tried to merge the extension into current trunk. After succesfull merging I had to
 `sudo apt remove python-xmldict`
 `pip2 install --user python-xmldict`
to get the current version 0.11 of xmldict and the extension running. 

We really need a install check and version check with proper error messages.
Maybe even an error message, if running on win or macos.

Installation requirements should be mentioned in the README file.

myselfhimself:
Indeed, by the way in Python, the standard is to have dependencies listed
in a requirements.txt file, which pip can install with -r filename

swzCuroverse:
I am stuck in the same place as myselfhimself where I hit the apply button and can't get any further.  I am running Ubuntu.  I tried both in python2 and python3 (python3 errs due to some library incompatibilities). Not sure how to further debug.

myselfhimself:
My Silhouette is back in its packaging box due to relocation. I do not know
when I will reset it up and restart working with it sorry :-/

ne 28. 4. 2019 v 13:40 odesílatel SWZ <notifications@github.com> napsal:

Issue #47, opened 2017-07-15
debian installation instructions: fetching release key: 404 error


AVHon:
Running the following line:
```wget http://download.opensuse.org/repositories/home:jnweiger:fablab/xUbuntu_14.10/Release.key -O - | sudo apt-key add -```
on Debian Jessie 8 fails. Specifically, wget 404's on the link. The proper version should be:
```wget https://download.opensuse.org/repositories/home:jnweiger:fablab/xUbuntu_14.10/Release.key -O - | sudo apt-key add -```

EtherGraf:
Just tested it on my ubuntu. 
Both lines fails on ubuntu with 404. So I don't think the https protocol is the only reason.

EtherGraf:
For me 
`http://download.opensuse.org/repositories/home:/jnweiger:/fablab/xUbuntu_14.04/Release.key`
is the correct path and works both with http and https.
Pay attention to the semicolons.

jnweiger:
The README.md is a mess. Sorry for that.
Ive added references to github releases there but failed to remove the old references to the opensuse build service. The build project home:jnweiger:fablab no longer exists.

PR #48, opened 2017-07-15
fix for #47: wget key 404s

AVHon:
This change just changes the "wget key" line from an http link to an https.


EtherGraf:
@jnweiger:
Shouldn't this pull request be closed as the build service doesn't exist any more?

EtherGraf:
#47 is closed already. 
The build project home:jnweiger:fablab no longer exists.

PR #56, opened 2017-12-28
properly calculate transforms when selecting objects in groups

### LINKED ISSUE #43: when specific objects are selected, parent transforms are not applied
Let's say I have a document with red and blue lines.  I want to cut the red ones and score the blue ones.  I'll select a red line, do *Edit* -> *Select Same* -> *Stroke Color* to select all the red ones.  If some of the lines are in groups, all hell breaks loose.

The problem is that `recursivelyTraverseSvg` is being passed the selected paths directly, not the groups they're in.  If a line is in a group with a transform, that transform is not applied to the line.  Stuff ends up all over the place.

I'm not sure what the best solution is.  [inkscape-embroidery](https://www.github.com/lexelby/inkscape-embroidery) gathers transforms differently, using `simpletransform.composeParents()`.  Perhaps that would help?

lexelby:
fixes #43 

cc @EtherGraf

### DIFF
--- a/sendto_silhouette.py
+++ b/sendto_silhouette.py
@@ -129,6 +129,8 @@ import silhouette.StrategyMinTraveling
 N_PAGE_WIDTH = 3200
 N_PAGE_HEIGHT = 800
 
+IDENTITY_TRANSFORM = [[1.0, 0.0, 0.0], [0.0, 1.0, 0.0]]
+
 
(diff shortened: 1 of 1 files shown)


EtherGraf:
Ok, have tested this branch and can verify it solves the problem. 
Not found any side effects up to now.

Thanks for your work.

PR #60, opened 2018-03-07
Update Graphtec.py for use with windows. 

sctv:
Added a test for Windows, as Windows does not seem to detect the usb.version , gives attribute error. 
If it passes the test, the plotter works fine. Other tests of pyusb work, pyusb is installed.
I am a beginner programmer, so there may be a better way of doing this.


sctv:
Changes are already in #62

sctv:
sctv:patch-2 has all of the changes.

PR #64, opened 2018-05-16
Inkscape Template for US Letter in Landscape

KedneckInc:
# DESCRIPTION

This file is a modified copy of the 'default.svg' file.  It is
configured for US Letter paper in landscape mode, for the Silhouette
Cameo 3 with cutting mat, by the following additions:

1) The overall page size has been increased by ~20mm in both directions.

   This avoids issues where the software truncates any line/cut that
   extends beyond a certain point.

2) Guides have been added which mark the 'safe' boundaries for cutting.

   The guides have been moved in from the upper left by 5mm, allowing
   a safety zone against misalignment of paper on the cutting mat.
   The lower right guides have been moved in from the edges by 25mm,
   allowing for the 20mm increase, and an additional 5mm for the
   safety zone.

3) A grid has been pre-defined, in milimeters, major lines every 10mm.

4) Both display and document units are set to milimeters.

# LICENSE

This file is under the CC BY 4.0 license as defined by:

  https://creativecommons.org/licenses/by/4.0/

### DIFF
--- a/examples/default_us_letter_landscape_cameo_3.svg
+++ b/examples/default_us_letter_landscape_cameo_3.svg
@@ -1,0 +1,123 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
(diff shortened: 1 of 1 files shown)


KedneckInc:
@EtherGraf I took your suggestion on setting the media size larger, and mixed it in with a web page that described how to go about setting up a personalized default template in inkscape, and came up with this file.  I suggest that it would make it easier for new users to have at least one of these templates for working with the Cameo 3.

I considered updating the README.md to contain a description of the file, but that would require descriptions of the other examples as well.  I'm willing to do them, but I can't guarantee that I'd get them completely right.

Let me know what you think.

EtherGraf:
Sorry, I am at a hurry at the moment and have no access to computer the next week. 

Thank you very much for your valuable work. I will support and merge it. Please continue your work.
I had the same thoughts, but not found time to investigate further.

As a final solution I think, the code should observe both the document size and the different device limits, but this will be a bigger task.

Issue #49, opened 2017-07-31
default DPI changed in Inkscape 0.92+


mehtank:
[ see: http://wiki.inkscape.org/wiki/index.php/Release_notes/0.92#Important_changes ]

The way units are handled changed in Inkscape 0.92, and it appears to have broken this plugin, but I haven't quite been able to figure out how.  Sometimes things get cut with the correct dimensions, sometimes they're cut too small at 90/96 scale, and sometimes too big at 96/90 scale.  Perhaps there's some assumption on the units that is not valid with the new Inkscape?

amittel:
Hi!
Just using this plugin. I use `mm` as unit and had no problems. But I found a conversion function with a [hardcoded dpi](https://github.com/fablabnbg/inkscape-silhouette/blob/master/sendto_silhouette.py#L135).
If you change it to 96dpi it should work. Just open it in an editor and change it and should work.

jnweiger:
Yes, this dpi change needs to be supported by the plugin. I think the px2mm() function you found is the right place for this. Code needs to be added there, to do effectively the same that we do in 

 https://github.com/fablabnbg/inkscape-paths2openscad/blob/master/paths2openscad.py#L511

A different workaround with Inkscape 0.92 should be 
 -> Extensions -> Document -> DPI 90 to 96

lexelby:
From what I can tell, loading an old file in inkscape 0.92 and selecting "set viewbox" works to correct the dpi change, but *only* if the old document's dimensions were specified in mm or inches.  That is, the `<svg>` tag's `width` and `height` must include units in them.

The reason this works is that this plugin will first interpret those dimensions (incorrectly) in 90 DPI due to `getLength()`, and then interpret the viewbox in pixels, resulting in an identity transform as the `docTransform`.  Then it will interpret all pixel measurements in path objects as if they were specified in 90 DPI, due to `px2mm()`.

In essence, the two places in which the plugin hardcodes 90 DPI cancel each other out!  The viewbox tells inkscape to scale things to work with its 90 DPI, and dimensions are correctly reflected in the UI.  The plugin essentially ignores the viewbox and works in 90 DPI, and that works out too.

[...]

PR #67, opened 2018-06-09
Changing 90 to 96 dpi to follow Inskapce's behavior

petermora:


EtherGraf:
I expect, this commit will break compatibility of the extension with old 0.91 programm/files.
See some discussion at #49.

EtherGraf:
An universal solution for all versions will need more work, but as a first step I accept this pull request. At least with the current 0.92 it should work now correctly.

PR #69, opened 2018-07-07
Add file output option for remote cutting

neocogent:
Sometimes it would be nice to cut remotely or across a network. This set of changes adds the ability to "cut to file" so you can dump it to the cutter elsewhere. It works by adding a dummy PrnFile class which fakes and removes the status commands while writing actual cut data to file. An option for filepath is provided.

Note it does not create a GSD file suitable for putting on an SD card unfortunately. That would have been nice too but the raw cut data is not the same as the GSD format.

You can send the prnfile data using your typical cat or echo command (on linux, probably different on other OS). eg. `cat test.cut > /dev/usb/lp1` (that worked for me once I added myself to lp group).

My main reason for wanting this was so I can put the cutter in the "noisy room" with 3D printer hanging off a raspberry pi but still do design work on my laptop in a quieter room. But it will work also when you want to save cut jobs for later use and just send them rather than opening inkscape. So it is nice for automated repetitive tasks as well.

It would be nice to add a file chooser for selecting output filename but I stopped short due to work required to research and figure that out.

Hope this is as useful for others as for me. Cheers.


t0b3:
resolved with PR #140

PR #85, opened 2019-07-25
merge jnweiger and olegdeezus

jnweiger:


jnweiger:
Damned, the previous merge already polluted fablabnbg/master, while using the github web merge tool, to merge fablabnbg into jnweiger. Github tools are insane. Merging back to make both master branches fablabnbg and jnweiger finally equal and containing all from olegdeegaz (and from several others what was accumulated in jnweiger/master during the last two years.

Sorry for the monster commits here.

PR #94, opened 2020-04-21
Add Python 3 compatibility

t0b3:
this PR paves the way to inkscape 1.0
works like a charm with python3.6 + inkscape-0.92.5 and inkscape-1.0_rc1
while still working with python2.7 + inkscape-0.92.4

### DIFF
--- a/eggbot_hatch.py
+++ b/eggbot_hatch.py
@@ -319,7 +319,7 @@ def interstices( P1, P2, paths, hatches, margin=0 ):
         # Now, entries with even valued indices into sa[] are where we start
         # a hatch line and odd valued indices where we end the hatch line.
         for i in range( 0, len( sa ) - 1, 2 ):
-                if not hatches.has_key( sa[i][1] ):
+                if sa[i][1] not in hatches:
                         hatches[sa[i][1]] = []
                 x1 = P1[0] + sa[i][0] * ( P2[0] - P1[0] )
                 y1 = P1[1] + sa[i][0] * ( P2[1] - P1[1] )
@@ -859,7 +859,7 @@ class Eggbot_Hatch( inkex.Effect ):
 
                                 pass
 
-                        elif not isinstance( node.tag, basestring ):
+                        elif not isinstance( node.tag, str ):
 
                                 pass
 
@@ -1022,7 +1022,7 @@ class Eggbot_Hatch( inkex.Effect ):
                 for key in self.hatches:
                         path = ''
                         direction = True
-                        if self.transforms.has_key( key ):
+                        if key in self.transforms:
                                 transform = inverseTransform( self.transforms[key] )
                                 # Determine the scaled stroke width for a hatch line
                                 # We produce a line segment of unit length, transform
--- a/examples/testcut_sq_tr.py
+++ b/examples/testcut_sq_tr.py
@@ -26,8 +26,8 @@ def write_progress(done, total, msg):
 
 dev = SilhouetteCameo(progress_cb=write_progress, no_device=False)
 state = dev.status()    # hint at loading paper, if not ready.
-print "status=%s" % (state)
-print "device version: '%s'" % dev.get_version()
+print("status=%s" % (state))
+print("device version: '%s'" % dev.get_version())
 
 dev.setup(media=132, pen=False, pressure=1, speed=10)
 dev.plot(pathlist=test_sq_tr, offset=(30, 30))
--- a/examples/testcut_sq_tr.py
+++ b/examples/testcut_sq_tr.py
@@ -4,6 +4,9 @@
 # as our triangle is not a simple triangle, but a more complex
 # Arrow head. The outer square is an ordinary 1cm x 1cm square.
 #
+
+from __future__ import print_function
+
 import sys
 sys.path.append('/usr/share/inkscape/extensions')
 sys.path.append('/home/jw/src/github/fablabnbg/inkscape-silhouette')
--- a/misc/arrow_test.py
+++ b/misc/arrow_test.py
@@ -6,6 +6,8 @@
 #
 # Requires: python-usb  # from Factory
 
+from __future__ import print_function
+
 import time, sys, os
 sys.path.append(os.path.dirname(os.path.abspath(__file__)) + '/..')      # make it callable from anywhere
 from silhouette.Graphtec import SilhouetteCameo
--- a/misc/arrow_test.py
+++ b/misc/arrow_test.py
@@ -12,11 +12,11 @@ from silhouette.Graphtec import SilhouetteCameo
 
 # coordinates in mm, origin int top lefthand corner
 arrow1 = [ (0,5), (21,5), (18,0), (31,10), (18,20), (21,15), (0,15), (3,10), (0,5) ]
-arrow2 = map(lambda x: (x[0]+263, x[1]+0), arrow1)
-arrow3 = map(lambda x: (x[0]+30, x[1]+0), arrow1)
+arrow2 = [(x[0]+263, x[1]+0) for x in arrow1]
+arrow3 = [(x[0]+30, x[1]+0) for x in arrow1]
 
 dev = SilhouetteCameo()
-print "status=%s" % dev.wait_for_ready(20)
+print("status=%s" % dev.wait_for_ready(20))
 
 tmp_fwd=85	# enough to show 20mm of the latest drawing on the far side of the device.
 
@@ -32,4 +32,4 @@ for i in range(8):
   # we can workaround by sending a dummy job.
   bbox = dev.plot(pathlist=[ [(0,0),(0,0)] ], offset=(0,0))
   dev.wait_for_ready()
-print bbox
+print(bbox)
--- a/misc/endless_clock.py
+++ b/misc/endless_clock.py
@@ -9,6 +9,8 @@
 # Modelled after https://github.com/rougier/freetype-py/blob/master/examples/glyph-vector.py
 # first working draft, doing 2 things at a time.
 
+from __future__ import print_function
+
 import sys,time,gtk
 from goocanvas import *
 import cairo,random
--- a/misc/endless_clock.py
+++ b/misc/endless_clock.py
@@ -145,12 +145,12 @@ def scale_up(win, ev):
   if   chr(ev.keyval) == '+':  canvas.set_scale(s*1.2)
   elif chr(ev.keyval) == '-':  canvas.set_scale(s*.8)
   else: gtk.main_quit()
-  print canvas.get_scale()
+  print(canvas.get_scale())
 
 def button_press(win, ev):
   win.click_x = ev.x
   win.click_y = ev.y
-  print win.click_x, win.click_y
+  print(win.click_x, win.click_y)
 
 def button_release(win, ev):
   win.click_x = None
--- a/misc/font_metrics_bug.py
+++ b/misc/font_metrics_bug.py
@@ -7,6 +7,8 @@
 # https://bugzilla.novell.com/show_bug.cgi?id=820593
 # https://bugzilla.gnome.org/show_bug.cgi?id=700664
 
+from __future__ import print_function
+
 import gtk
 from goocanvas import *
 import cairo
--- a/misc/font_metrics_bug.py
+++ b/misc/font_metrics_bug.py
@@ -16,12 +16,12 @@ def scale_up(win, ev):
   if   chr(ev.keyval) == '+':  canvas.set_scale(s*1.2)
   elif chr(ev.keyval) == '-':  canvas.set_scale(s*.8)
   else: gtk.main_quit()
-  print canvas.get_scale()
+  print(canvas.get_scale())
 
 def button_press(win, ev):
   win.click_x = ev.x
   win.click_y = ev.y
-  print win.click_x, win.click_y
+  print(win.click_x, win.click_y)
 
 def button_release(win, ev):
   win.click_x = None
--- a/misc/silhouette_cut.py
+++ b/misc/silhouette_cut.py
@@ -34,7 +34,7 @@ dumpdata=dev.load_dumpfile(args.dumpfile)
 dev.wait_for_ready()
 meta = dev.plot(pathlist=[], bboxonly=args.bbox, no_trailer=True,
                 mediawidth=args.width, mediaheight=args.height, offset=(args.xoff,args.yoff))
-print meta
+print(meta)
 
 dev.wait_for_ready()
 cmd_list = dev.plot_cmds(dumpdata, meta['bbox'], args.xoff, args.yoff)
--- a/misc/silhouette_cut.py
+++ b/misc/silhouette_cut.py
@@ -5,6 +5,8 @@
 #
 # Requires: python-usb  # from Factory
 
+from __future__ import print_function
+
 import re,time,sys,string,argparse
 
 sys.path.extend(['..','.'])	# make it callable from top or misc directory.
--- a/misc/silhouette_move.py
+++ b/misc/silhouette_move.py
@@ -5,6 +5,8 @@
 #
 # Requires: python-usb  # from Factory
 
+from __future__ import print_function
+
 import time,sys,string
 
 sys.path.extend(['..','.'])	# make it callable from top or misc directory.
--- a/misc/silhouette_move.py
+++ b/misc/silhouette_move.py
@@ -17,6 +17,6 @@ if len(sys.argv) > 1:
   feed_mm = string.atof(sys.argv[1])
 
 if not feed_mm: 
-  print "Usage: %s [PAPER_FORWARD_MM]" % sys.argv[0]
+  print("Usage: %s [PAPER_FORWARD_MM]" % sys.argv[0])
 
 dev.move_origin(feed_mm)
--- a/misc/test_drive.py
+++ b/misc/test_drive.py
@@ -146,7 +146,7 @@ cut = cut_vertical_zigzag + [[(10,0.2),(12,0.3),(11.6,0.8)]]
 if len(sys.argv) > 1:
   str=open(sys.argv[1]).readlines()
   cut=eval("".join(str), {}, {})
-  print "loading from", sys.argv[1]
+  print("loading from", sys.argv[1])
 
 
 def print_pdf(c):
@@ -168,7 +168,7 @@ def print_pdf(c):
   # ctx.fill()
   c.render(ctx)
   ctx.show_page()
-  print "output.pdf written"
+  print("output.pdf written")
 
 def key_press(win, ev, c):
   new_idx = None
@@ -192,7 +192,7 @@ def key_press(win, ev, c):
       a.remove()
     c.arrow = []
   else:
-    print c.get_scale()
+    print(c.get_scale())
 
   if new_idx is not None:
     jumpto = True
@@ -216,7 +216,7 @@ def key_press(win, ev, c):
                 end_arrow=True, arrow_width=3, arrow_tip_length=2, arrow_length=2, 
                 stroke_color_rgba=0x00000033))
       c.cursor.animate(cx,cy, 1, -360., absolute=True, duration=150, step_time=30, type=0)
-    print new_idx, c.points[c.cursor_idx][0], c.points[c.cursor_idx][0].att()
+    print(new_idx, c.points[c.cursor_idx][0], c.points[c.cursor_idx][0].att())
   else:
     c.cursor.stop_animation()
 
@@ -288,7 +288,7 @@ def main ():
     if len(canvas.points) <= 1:
       rect = Rect(parent=root, x=1, y=1, width=3,  height=2,
                   fill_color = '#77ff77', stroke_color = 'black', line_width = .01)
-      print "Dummy rectangle drawn. No points in canvas."
+      print("Dummy rectangle drawn. No points in canvas.")
     else:
       cursor_p = Points([(-0.5,0),(0,0.5),(0.5,0),(0,-0.5),(-0.5,0)])
       canvas.cursor = Polyline(parent=root, points=cursor_p, line_width=0.05, stroke_color="green", fill_color_rgba=0x77ff7777)
--- a/misc/test_drive.py
+++ b/misc/test_drive.py
@@ -21,6 +21,8 @@
 #
 # Requires: python-goocanvas
 
+from __future__ import print_function
+
 import sys
 import gtk
 from goocanvas import *
--- a/sendto_silhouette.py
+++ b/sendto_silhouette.py
@@ -87,6 +87,8 @@
 #                          merge from fablabnbg
 # 2019-08-03, jw, v1.22 - added a copy of pyusb-1.0.2 as a fallback on any platform.
 
+from __future__ import print_function
+
 __version__ = '1.22'	# Keep in sync with sendto_silhouette.inx ca line 79
 __author__ = 'Juergen Weigert <juergen@fabmail.org> and contributors'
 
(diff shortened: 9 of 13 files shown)


pierred35:
Thank you for this improvement. 
I have Python3.8 (with pyusb and pywinusb installed using pip) + inkscape-0.92.5
I've copy your 3 elements (silhouette folder, sendto_silhouette.inx and sendto_silhouette.py) in the Inkscape\share\extensions folder
Now in Inkscape, when I try to cut something, the box "send to silhouette" appears quarter of a second and dissapears directly and nothing happen on the Cameo. 
Do you have an idea where could be the problem? Thanks for your help...

t0b3:
@pierred35 we need more information to tackle your situation. Could you start inkscape from a cmd-Shell, observe the messages written to shell and report them back? You may also proceed as explained here to debug on windows https://superuser.com/a/1472274

jnweiger (changes requested):
Thanks for pulling forward!
Thanks for taking care of py2/py3 and inkscape 0.9x/1.0 compatibility. That's exactly how I want the code to be.

* [x] Files in examples and misc should also have a `from __future__ import print_function`, please.
* [x] Graphtec.py read() suggest to change `if isinstance(data, str):` to `if isinstance(data, (str, bytes)):` to be safer. But please test, I am not sure.
* [x] Graphtec.py: plot_cmds() please don't drop the enable_sw_clipping test here. Needs to be discussed in a separate PR, please. That switch was added for cameo3 in b87652149609a8bedcdc034dd0e3e5cfe38382fe

I'd like to wait for @EtherGraf to also leave a comment.

jnweiger (approved):
That's great now. Thank you!

EtherGraf:
Had just a quick look at it. I am not an experienced python programmer, but so far this patch looks good for me. Would like to test, but have no running inkscape 1.0 instance on linux at hand.

I had started a private python3 branch myself some time ago, but this one looks better than my first steps.

jnweiger:
> Had just a quick look at it. I am not an experienced python programmer, but so far this patch looks good for me. Would like to test, but have no running inkscape 1.0 instance on linux at hand.

I'll fire up 0.92.5 tomorrow and 1.0rc and see how that works.

pierred35:
@t0b3 still looking how to do to have a log file. Sorry, I'm not an expert. 
in dos-cmd, I go in the Inkscape folder then use : inkscape SET INKSCAPE_DEBUG_LOG=c:\temp\testdebug.xml but I have a message error "parser error : document is empty". 
I'm lost, I don't know how to do. Thank you for your help.

pierred35:
[testdebug.zip](https://github.com/fablabnbg/inkscape-silhouette/files/4560037/testdebug.zip)
Finally, I found the way how to creat a log file with
SET INKSCAPE_DEBUG_FILTER=CORE,XML,SPOBJECT,INTERACTION,CONFIGURATION
Let me know if it's enouth, thanks for your help.

t0b3:
@pierred35 try again with SET INKSCAPE_DEBUG_FILTER=[EXTENSIONS](https://gitlab.com/inkscape/inkscape/-/blob/master/doc/LOGGING.txt#L15) to see what's going on there.

NB: Do you know whether it used to work with python2 on your machine? Do you have both python versions in parallel available to switch and test?

pierred35:
@t0b3 
If I only SET INKSCAPE_DEBUG_FILTER=EXTENSION, my log file is empty.
My configuration : 
- Silhouette Cameo 1
- only Python3.8 (with pyusb and pywinusb installed using pip) 
- inkscape-0.92.5
- installed the new driver with Zadig-2.5
- from "inkscape-silhouette-master", I have copied "silhouette" folder and 2 files "sendto_silhouette.py and .inx" in "C:\Programs\Inkscape\share\extensions"
- form "pyusb-1.0.2", I have copied folder "usb" to "C:\Progams\Inkscape\lib\python2.7\site-packages" --> is it the good folder ? (I also tried "C:\Progams\Inkscape\lib\python3.8\site-packages"
I'm not certain for that step. 
Thanks a lot for your help.

EtherGraf:
To get more basic diagnostic  information, you can start the plotter at commandline without inkscape.

Open a cmd window.
Enter
C:\programms\[...]\python.exe sendto_silhouette.py [yourdoc].svg
(Adjust paths as needed.)

What does happen?

pierred35:
I've copied an example file in c:\temp\testcut.svg
In cmd windows, c:\program Files (x86)\python\python.exe sendto_silhoette.py .svg
--> Python.exe: can't open file
c:\program Files (x86)\python\python.exe sendto_silhoette.py c:\temp\testcut.svg
--> Python.exe: can't open file

Are you talking about python or Inkscape? 
In cmd windows, c:\program files\inkscape\inkscape.exe sendto_silhoette.py .svg or c:\program files\inkscape\inkscape.exe sendto_silhoette.py c:\temp\testcut.svg
--> message error : can't load the file 
Thanks for trying to solve the problem

EtherGraf:
I am talking about the python.exe. You do not need to start inkscape for this basic tests.
I am on linux here, so I don't know the exakt windows paths. You have to check by yourself.

Also I missed the u in my example line: sendto_silhoette.py != sendto_silhouette.py 

Find the path to python.exe in the programm files.
Find the path to the extension.

Then use the single line
`c:\program Files (x86)\python\python.exe [path to extension]\sendto_silhouette.py c:\temp\testcut.svg`


This should cut the content of testcut.svg.

pierred35:
Thank you @EtherGraf, we progress. 
Sorry, I'm not at all an expert in linux, cmd line, and so. 
SO, using you cmd line, lxml was missing. I've installed it with pip. This is ok now. 
now, when I launch the same cmd line, I have that
msg_error
My Cameo is ready. 
What to do now? 
Thank you all so much for your help

EtherGraf:
From the error message above I can see, that there is some problem in the communication with the plotter. The data could not be sent to the plotter.

I suspect the pyusb, but have no idea how to go on from here.
Any windows user wo can give some hint?

EtherGraf:
@jnweiger 
In the recent revision 4333fd1 line 430 in Graphtec.py gots changed from

s.write("\x1b\x05")
resp = "None\x03"

to

s.write(b"\x1b\x05")
resp = b"None\x03"`

Could that possibly lead to a problem?

EtherGraf:
@t0b3 @jnweiger 
I think the problem is caused by a wrong data type given to s.write() in Graphtec.py
If I assume the s.dev.write in line 332 will throw an exception, then this exception will get caught silently. After 5 tries the above error message will appear.

t0b3:
@pierred35 open python.exe and give as input
```
import usb
usb.version_info
VENDOR_ID_GRAPHTEC = 0x0b4d
dev = usb.core.find(idVendor=VENDOR_ID_GRAPHTEC)
dev
dev.set_configuration()
dev.set_interface_altsetting()
print("\x1b\x05")
dev.write(0x01, "\x1b\x05", timeout=10000)
print(b"\x1b\x05")
dev.write(0x01, b"\x1b\x05", timeout=10000)
```
and post back your output

jnweiger:
> @t0b3 @jnweiger
> I think the problem is caused by a wrong data type given to s.write() in Graphtec.py
> If I assume the s.dev.write in line 332 will throw an exception, then this exception will get caught silently. After 5 tries the above error message will appear.

That might be the case. We should see the exception, to be sure. 

But I think we still miss a python3 usb module.
@pierred35 tried moving an usb folder from "pyusb-1.0.2" to different locations. That is pretty certain a python2 module. We need python3 code now. I currently don't know if a pyusb windows package or something similar exists for python3.

pierred35:
python_msg
@t0b3 , I've tried your code: at print("\x1b\x05") line, my Cameo turns the roller and never stop. I've hit the power button after 5min. Good news? 
@jnweiger, I've moved the usb forder from "C:\Program Files (x86)\Python\Lib\pyusb-1.0.2" to "c:\temp\usb" but still the same problem. 
python_msg 2

jnweiger:
That is a great test, @pierred35 Thanks! (I wish i had a cameo and a windows box myself)
`dev.write(0x01, "\x1b\x05", timeout=10000)` responds with 4. Does that mean 4 bytes got send? We only wanted to send two. 
`dev.write(0x01, b"\x1b\x05", timeout=10000)` responds with 2. So I believe that is what we want.
(In my world nothing exists except bytes. I am a senior C-hacker. Perl and python's attempts to do every I/O with a lot  of re-encoding in the fly was always horrors to me. So happy to have proper bytes support back in python3 now)

The dll error in the second screenshot, I don't understand. I can still assume it is an artefact of mixing python2 code with python3.

pierred35:
@jnweiger thank you for your support
python_msg 3
What do you want I test now? Any idea? 
Thanks, and sorry to not be able to help more or diagnose something by myself, I don't know nothing about programming.

t0b3:
@jnweiger [pyusb](https://github.com/pyusb/pyusb) ist [python3](https://pypi.org/project/pyusb/) compatible since many years.
@EtherGraf this case shows we should keep the byte data type b'...' to not end up with encoding issues.
@pierred35 your [screenshot](https://user-images.githubusercontent.com/63608503/80859294-d0bc3b00-8c5f-11ea-90ce-86f5ae93118b.png) looked quite promising. go back to that pyusb. next try
```
import sys
print(sys.version)
import os
print(os.getcwd())
# sys.path.append(os.path.dirname(os.path.abspath(__file__)) + '/pyusb-1.0.2')
print(sys.path)
import usb
print(usb.__version__)
VENDOR_ID_GRAPHTEC = 0x0b4d
dev = usb.core.find(idVendor=VENDOR_ID_GRAPHTEC)
dev.set_configuration()
dev.set_interface_altsetting()
dev.write(0x01, b"\x1b\x05", timeout=10000)
resp = dev.read(0x82, 64, timeout=5000)
print(resp)
print(resp.tostring().decode())
dev.write(1, b"FG\x03")
resp = dev.read(0x82, 64, timeout=5000)
print(resp)
print(resp.tostring().decode())
```

pierred35:
python_msg 4
What I should enter on the next line?

t0b3:
@pierred35 retry without this line

pierred35:
image
after that line, I have always the same message "Traceback..."

t0b3:
@pierred35 this looks odd but not bad. continue execution to the end...

pierred35:
image

t0b3:
```
import usb
VENDOR_ID_GRAPHTEC = 0x0b4d
dev = usb.core.find(idVendor=VENDOR_ID_GRAPHTEC)
print(dev)
```

pierred35:
image

EtherGraf:
Do you have another software running in parallel, e.g silhouette studio, which could block this usb device?

pierred35:
not at all. I just restart my computer to be certain and I have the same result.

t0b3:
@pierred35 this indicates a problem with your pyusb installation oder an usb communication problem with your plotter. 
It would be interesting to see what happens if you delete all the *.pyc in your pyusb folder and retry again...

pierred35:
ok, so, inside C:\Program Files (x86)\Python\Lib\pyusb-1.0.2\usb\  I have delete all the *.pyc that were all in the _pycache_ folder and test again

I disconect the usb device and plug it in another usb port and now I have somethin
image

pierred35:
We are progressing !!!
image

t0b3:
I fixed some typos in the instructions above. give it a try again...

pierred35:
sorry where?

pierred35:
image

t0b3:
there were missing brackets after `tostring`

```
import sys
print(sys.version)
import os
print(os.getcwd())
# sys.path.append(os.path.dirname(os.path.abspath(__file__)) + '/pyusb-1.0.2')
print(sys.path)
import usb
# print(usb.__version__)
VENDOR_ID_GRAPHTEC = 0x0b4d
dev = usb.core.find(idVendor=VENDOR_ID_GRAPHTEC)
dev.set_configuration()
dev.set_interface_altsetting()
dev.write(0x01, b"\x1b\x05", timeout=10000)
resp = dev.read(0x82, 64, timeout=5000)
print(resp)
print(resp.tostring().decode())
dev.write(1, b"FG\x03")
resp = dev.read(0x82, 64, timeout=5000)
print(resp)
print(resp.tostring().decode())
```

pierred35:
image

t0b3:
@pierred35 communication with cameo looks good. pyusb looks good. now proceed with instructions above [python.exe sendto_silhouette.py test.svg](https://github.com/fablabnbg/inkscape-silhouette/pull/94#issuecomment-622422934)

pierred35:
image

pierred35:
if I return in the python cmd and test again :
image

pierred35:
hey, so I close the python cmd box and try again in the windows cmd box and it's work !!!

... but it doesn't work directly from Inkscape... where can be the problem? thanks for your help, I really appreciate

pierred35:
If I modify the testcut.svg file with Inkscape, and launch again the cmd in the windows cmd box, the Cameo move according to my drawing! This is great!
But using the extension inside Inkscape, that doesn't work. So I think the problem is the output from Inkscape. It doesn't send good settings to the Cameo.
Can we test with other parameters?

t0b3:
so this sounds like a communication issue somewhere between inkscape and sendto_silhouette.py or maybe your inkscape installation uses a different python version or python context?

you can use cmd to 

1. configure relevant variables [PATH, PYTHONPATH, etc](https://docs.python.org/3/using/windows.html#configuring-python).
2. reproduce above test with `python sendto_silhouette.py test.svg` to ensure your setting works
3. call `inkscape.exe` from that very same cmd and review the result
if that procedure fails again it might be because there are maybe other python installation on your system, or within your inkscape installation. scan for python.exe in inkscape folder or entire c:\

in case there exist other python installations, ensure they also have the relevant pip modules (pyusb, lxml, etc.) installed within their folders.

pierred35:
I'm sorry but I'm very limited in that domain. I feel like I'm going in circles
Here is what I get, trying to "Setting environment variables"
image
What should I put in the line C:\>set PYTHONPATH=%PYTHONPATH%;C:\My_python_lib ? because c:\My_python_lib doesn't exist
My only one Python is installed here : C:\Program Files (x86)\Python\
and Python lib here : C:\Program Files (x86)\Python\Lib\
I'm sorry, it should be something very easy for you but difficult for me...
image

t0b3:
@pierred35 open explorer. check out your inkscape folder. watch out for python.exe

pierred35:
Here is the result when I'm looking for python.exe on all my hard drive:
image
So no python in the Inkscape folder

t0b3:
`echo %PATH%`

pierred35:
image
image

t0b3:
`echo %PYTHONPATH%`

pierred35:
image

t0b3:
```
cd ___path-to-your-inkscape-installation___
dir
```

pierred35:
image
image
image

t0b3:
what happens if you start from right that directory [this command](https://user-images.githubusercontent.com/63608503/80864813-69b17d00-8c85-11ea-8f9a-83494ebf5863.png)?

pierred35:
image
same problem

t0b3:
> hey, so I close the python cmd box and try again in the windows cmd box and it's work !!!
> 
> ... but it doesn't work directly from Inkscape... where can be the problem? thanks for your help, I really appreciate

what did you do different this time? it seemed to work earlier, right?

pierred35:
nothing different... I don't understand what happen...

t0b3:
is it the different folder you start?
maybe unplug and replug your plotter helps?

pierred35:
So I redo the procedure 
image
and it's work now...

t0b3:
retry from inkscape dir works now? or maybe it's what matters?

pierred35:
image
python.exe not recognize as internal command...

t0b3:
> image
> same problem

but so you did something different than before, right?

pierred35:
image
and so on...

pierred35:
So, when I open windowd cmd, 
- if I try 
image
that doesn't work
- now if I go in the python cmd, do your code
"import sys, import os, import usb..."
i have the "CAMEO 1.10" at the end,
then close the python cmd and try again the code, that works on the CAMEO

t0b3:
... and then next try to start from right that window `"C:\Program Files\Inkscape\inkscape.exe"`

pierred35:
I just do it but nothing happen, Cameo doesn't move...
You propose me to add Environment Variables. How to do that?

t0b3:
do other extensions in inkscape work after all? test some of them...

pierred35:
you're right, no extensions are working... same thing for all
maybe I should re-install Inkscape...

pierred35:
Ok, so I re-intall Inkscape with Python 2.7 and it's work now.
At the end, I have Python 2.7 installed in Inkscape folder and Python 3.8 install in the Program files folder. I don't know witch one is used by Silhouette extension to work. 
Thanks everybody for your time helping me. I really appreciate. 
Have a good night all

t0b3:
looks like your inkscape 0.92.5 uses the bundled python 2.7
I guess you messed something up with that integrated python

In case you want to test python 3 with inkscape without screwing things up, go for inkscape 1.0 rc1.

FlyingSamson:
When I try to use the plugin within Inkscape 0.92 using Python3.8 (set `python-interpreter="/opt/local/bin/python3.8"` in Inkscapes Preferences.xml) I obtain an error in `Graphtec.py:408` (inside the `read` routine) telling me that a 'str' object does not have a 'decode' attribute, which it indeed doesn't have anymore since Python3. I found this odd since the enclosing if statement does explicitly test whether we got a str object in order to then call a function that is not defined on one. Am I doing something wrong here in using Python3.8 with Inkscape 0.92.

I came across this while trying to make Bluetooth support work on all combinations of Inkscape (0.92 and 1.0) and Python (2 and 3), so it would be good to know if using this plugin with Python3 within Inkscape 0.92 isn't supposed to be supported anyway.

t0b3:
@FlyingSamson try again with Inkscape **0.92.5** as older versions won't be able to work with py3.

FlyingSamson:
Thank you for your fast reply. Actually it already is Inkscape 0.92.5:
```
$> inkscape --version
Inkscape 0.92.5 (2060ec1f9f, 2020-04-08)
```

t0b3:
@FlyingSamson you hit the point of testing I proposed to automate with in
PR 99
https://github.com/fablabnbg/inkscape-silhouette/pull/99

I didn't get the error you describe, neither on Linux, Mac, Windows,
whether with Python 2 nor 3 using any of the supported Inkscape Versions.

see the Travis CI Reports:
https://travis-ci.com/github/t0b3/inkscape-silhouette/builds/169285832

Any suggestions to enhance the tests in PR99 are highly appreciated.

FlyingSamson:
I'm not that familiar with python and even less so with the `subprocess` testing framework. But from what I understood from the `.travis.yml` file I presume the problem arises, as the CI-pipeline is using a dummy machine for testing. But the error occurs when calling `status()`. Here the complete stack trace that I receive from the plugin (sorry I forgot to include it in my first comment)
```
Traceback (most recent call last):
  File "sendto_silhouette.py", line 1238, in <module>
    e.affect()
  File "/opt/local/share/inkscape/extensions/inkex.py", line 289, in affect
    self.effect()
  File "sendto_silhouette.py", line 1021, in effect
    state = dev.status()    # hint at loading paper, if not ready.
  File "/Users/me/.config/inkscape/extensions/silhouette/Graphtec.py", line 438, in status
    resp = s.read(timeout=5000)
  File "/Users/me/.config/inkscape/extensions/silhouette/Graphtec.py", line 415, in read
    return str(data).decode()
AttributeError: 'str' object has no attribute 'decode'
```
`status()`, of course, never gets called when communicating with a dummy machine (when I use the dummy machine in Inkscape 0.92 and Python3.8 it also works).

Question is, whether this (`status()`) is the only case in which read is called with a `str` object. I would guess not.

I fear it would be difficult to include testing for such cases. The only option that comes to my mind, would be to write a 'real' dummy device, which does behave like a silhouette device and emulates the connection somehow, but I think that seems to be quite some work and then again, we couldn't really by sure whether we wrote the emulator correctly. One upside would be that the emulator of course would not be required to be written in both Python2 and Python3 (or Python at all), if it is run as a separate process on the CI-server (assuming that this would be possible to achieve in the travis configuration file).

t0b3:
@FlyingSamson I knew that kind of problems can arise and need to be
catched.
The tests you imagine are implemented in this commit
https://github.com/t0b3/inkscape-silhouette/commit/89c88302634ad3eac6ff57fc7ddbbdcb9671c11c

What looks odd to me is that the issue you mentioned does not reproduce
neither in Travis nor on my personal machine on py38 with Inkscape 0.92.5

I need to have a closer look into this as soon as I get back from holiday.

In the meantime you may have a look at this Travis test log and sort out
the differences to your setup
https://travis-ci.com/github/t0b3/inkscape-silhouette/jobs/327835582#L643

I.e. which code base do you use? The lines 415 and 438 mentioned don't make
sense with https://github.com/fablabnbg/inkscape-silhouette master branch.

Regards

FlyingSamson:
Ah I see. This library is pretty neat, I've got to remember this one.

Thank you for your pointers. I only now saw, that there is no pipeline for OSX with Python3 and Inkscape 0.92 (which is currently the configuration making the problems). Would it be possible to add such a pipeline?

I was using this [commit](https://github.com/FlyingSamson/inkscape-silhouette/commit/72f5e882655a88511c3340503d42085bd3905994) which includes only some extensions to make the plugin work with my Cameo4.

Until now I found out the following:
- `read` on Ubuntu uses the `read` routine of `dev`, which returns an array (on both Python2 and Python3), thus causing the `else` branch to fire which creates a `bytes` object on Python3 which has a `decode` method and a `str` object on Python2 which also has a `decode`method (in Python2)
- `read` on OSX uses the `bulkRead` routine of `dev`, which returns an `bytearray` (on both Python2 and Python3) , thus causing the `elif` branch to fire which creates a `str` object by calling `str()` on both Python2 and Python3, (however on Python2 the string contains, e.g., '0\x03' but on Python3 "bytearray(b'0\\x03')") and in case of Python3 the following `decode()` fails as there is no such method on `str`

Is there any reason why we should not also use `data.decode()` as in the `elif` branch (as in the `str`, `bytes`)?

t0b3:
@FlyingSamson thanks for letting me know types differ on Mac. Any
propositions how to fix it are highly welcome.

Nb: In case you know a method to automatically install this Inkscape
version on Mac with Travis, I'm very interested to extend the test setup to
include your case.

FlyingSamson:
How would I make a commit that tries to fix that problem, that will be checked by the pipeline? By forking from your repo and posting a merge request?


I'm not sure whether this is the only or the/a appropriate method of changing the python version used by Inkscape, but I just inserted the line `python-interpreter="/opt/local/bin/python3.8"` right below the line `id="extensions"` in my `preferences
.xml` file, so if I saw that correctly and the travis pipeline for OSX already has python3 installed (else it should be easily installable through brew), I would suggest to simply include a line like

```
sed -i.bak -E $'s#^([[:space:]]*)id="extensions"#&\\\n\\1python-interpreter="'"$(which python3)"'"#' ~/.config/inkscape/preferences.xml
```
Note that using the `.bak` after the -i option is mandatory because other than on Linux OSX's sed does not understand the empty string as 'do not make a backup at all'.

I hope that works in Travis as well. If not I will gladly help to make it work.

t0b3:
@FlyingSamson sorry I wasn't clear enough. I'm looking for a method to automatically install the exact Inkscape 0.92.5 version on OSX i.e. trough *brew*. 
Until now I didn't manage to have a reproducible OSX+**Inkscape092.5**+Python3 setup.

PR #96, opened 2020-05-01
fix py3 error

t0b3:
fix python3 issue: text can't be written to binary file

### DIFF
--- a/sendto_silhouette.py
+++ b/sendto_silhouette.py
@@ -1225,10 +1225,10 @@ if __name__ == '__main__':
 
         if len(sys.argv) < 2:
           # write a tempfile that is autoremoved on exit
-          tmpfile=tempfile.NamedTemporaryFile(suffix='.svg', prefix='inkscape-silhouette')
+          tmpfile=tempfile.NamedTemporaryFile(suffix='.svg', prefix='inkscape-silhouette', mode='w+t')
           sys.argv.append(tmpfile.name)
           print((sys.argv))
-          print('<xml height="10"></xml>', file=tmpfile)
+          tmpfile.write('<xml height="10"></xml>')
           tmpfile.flush()
           e.affect(sys.argv[1:])
           # os.remove(tmpfile.name)
--- a/sendto_silhouette.py
+++ b/sendto_silhouette.py
@@ -1227,7 +1227,7 @@ if __name__ == '__main__':
           # write a tempfile that is autoremoved on exit
           tmpfile=tempfile.NamedTemporaryFile(suffix='.svg', prefix='inkscape-silhouette', mode='w+t')
           sys.argv.append(tmpfile.name)
-          print((sys.argv))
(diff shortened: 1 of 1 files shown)


sendto_silhouette.py
```diff
@@ -1225,10 +1225,10 @@ def write_progress(done, total, msg):
 
         if len(sys.argv) < 2:
           # write a tempfile that is autoremoved on exit
-          tmpfile=tempfile.NamedTemporaryFile(suffix='.svg', prefix='inkscape-silhouette')
+          tmpfile=tempfile.NamedTemporaryFile(suffix='.svg', prefix='inkscape-silhouette', mode='w+t')
           sys.argv.append(tmpfile.name)
           print((sys.argv))
-          print('<xml height="10"></xml>', file=tmpfile)
+          tmpfile.write('<xml height="10"></xml>')
```
jnweiger, line 1231:
```suggestion
          print(b'<xml height="10"></xml>', file=tmpfile)
```
should do the trick too. (leaving the default mode="w+b"), doesn't it?

t0b3: ### Situation
unfortunately this does not work here with python3 due to some python internals.
```
import tempfile
tmpfile = tempfile.NamedTemporaryFile() # default mode='w+b' see https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile
print(b'<xml></xml>', file=tmpfile)
```
above code fails with
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.8/tempfile.py", line 474, in func_wrapper
    return func(*args, **kwargs)
TypeError: a bytes-like object is required, not 'str'
```
the deeper internals go beyond [tempfile.py](https://github.com/python/cpython/blob/master/Lib/tempfile.py#L475) where print() `*args` obviously become converted to `str` and won't fit the `mode='b'`.

Even `print(b"<xml></xml>", file=open("/tmp/test001.svg", "w+b"))` fails. This indicates it's a bad idea to use binary data with `print()`. Therefore I suggest to use the `write()` method.

### Purpose
the whole `if len(sys.argv) < 2` was introduced by [this commit](https://github.com/fablabnbg/inkscape-silhouette/commit/085def88c2c42296df99c69974321299f3509519#diff-ffbf6bd313ffa71608fa9f09fc9379a0R1068-R1075) in order to call `e.affect()` not without svg file.

t0b3:
@jnweiger wouldn't it be nice to have a test workflow checking the changes will run on all platforms with every python version available? - such checks could help to avoid checking in commits which break your fantastic code.

sendto_silhouette.py
t0b3: ### Solution
`write()` bytes to binary tmpfile
implemented in #99

t0b3:
closing in favor of #99

Issue #101, opened 2020-05-16
Inkscape 1.0 plots at wrong scale due to deprecates functions: simpletransform etc.


t0b3:
> Your wish let me work...but keep in mind: the extension is exactly the same, both Version use the SAME!! HOME partition.
> 
> Ubuntu Stdio 19.10 with inkscape 0.92.5 produces a nice output:
[silhouette.dump.zip](https://github.com/fablabnbg/inkscape-silhouette/files/4639163/silhouette.dump.zip)
> 
> On Ubuntu 20.04 / inkscape 1.0 python wont be found this way..
![Bildschirmfoto von 2020-05-16 22-26-16](https://user-images.githubusercontent.com/17491743/82129604-585b8b00-97c4-11ea-9c91-50a57a9868fd.png)
> 
> Called explicitly with python3 did run but output is different...
[terminal-output-20.04.zip](https://github.com/fablabnbg/inkscape-silhouette/files/4639202/terminal-output-20.04.zip)
> 
> and the dump
> [silhouette.dump-20.04.zip](https://github.com/fablabnbg/inkscape-silhouette/files/4639191/silhouette.dump-20.04.zip)

_Originally posted by @bianchifan in https://github.com/fablabnbg/inkscape-silhouette/issues/90#issuecomment-629704114_

@bianchifan I had a look at your dump files. they tell us quite a lot. the numer of points differ by 
* number of points
* scale
* placement of the letter 'o'

to evaluate dump files you may use [read_dump.py](https://github.com/fablabnbg/inkscape-silhouette/files/4639396/read_dump.py.txt)
![image](https://user-images.githubusercontent.com/5751047/82131437-754c8a00-97d5-11ea-8fc0-db72820234db.png)
this explains your issue. nb: the dump reproduces here. 
it looks like an error processing svg transform attribute with new inkex.

bianchifan:
OK..just two thoughts:
1st, unequal dump files due to different implementation and versions of python
2nd, your plots are nice but the output on my Cameo 2 rides other routes

Current (not really) working system: Ubuntu 20.04, upgraded from 19.10, roundabout 500 pragrams were purged during upgrade:((
python -> not installed, when trying to reinstall -> deprecated
python2 -> python 2.7
python3 -> python 3.8
Every call of "python" results in "not found", you must explicitly call python2 or python3.
When calling python3, python 3.8**!** is used.

Ubuntu Studio 10.10:
python -> python 2.7
[...]

Issue #90, opened 2020-02-16
update needed for Inkscape version 1.0


brynnd:
Hello,
Thank you for providing this extension for Inkscape users!

This is just to let you know that most 3rd party Inkscape extensions, like this one, probably will not work with the upcoming new Inkscape version, the long-awaited version 1.0.

Here is the info you need to update this extension, so that it will work with 1.0 and future versions.  
https://wiki.inkscape.org/wiki/index.php?title=Updating_your_Extension_for_1.0

If you have further questions, you can contact Inkscape developers via [mailing lists](https://lists.inkscape.org/postorius/lists/?all-lists), [forum](https://inkscape.org/forums/extensions/), or the [chatroom](https://chat.inkscape.org/channel/inkscape_extensions).

If you have already updated it, please disregard this message.

All best,
brynn

jnweiger:
Thanks for the reminder. We discussed that already on the last inkscape hackfests. I simply lack time to do it. Patches welcome.

t0b3:
@jnweiger PR #94 helps to solve this issue

iromero91:
For what it's worth, current master works perfectly on version 1.0 (tested on debian testing). Only issue is a huge amount of deprecation warnings when running the script.

t0b3:
just wondering what happens if you try this?
`sendto_silhouette.py --dummy=True examples/testcut_square_triangle.svg`
- any errors?

t0b3:
> For what it's worth, current master works perfectly on version 1.0 (tested on debian testing). Only issue is a huge amount of deprecation warnings when running the script.

@iromero91 just wondering what happens if you try this?
`sendto_silhouette.py --dummy=True examples/testcut_square_triangle.svg`
- any errors?

iromero91:
@t0b3 I get this when running from the console.
```
$ python sendto_silhouette.py --dummy=True examples/testcut_square_triangle.svg
sendto_silhouette.py:126: DeprecationWarning: inkex.localize was moved to inkex.localization.localize
  inkex.localize()	# 0.91
sendto_silhouette.py:270: DeprecationWarning: Effect.OptionParser or `optparse` has been deprecated and replaced with `argparser`.You must change `self.OptionParser.add_option` to `self.arg_parser.add_argument`; the arguments are similar.
  self.OptionParser.add_option('--active-tab', action = 'store', dest = 'active_tab',
[...]

PR #99, opened 2020-05-04
add automated testing of code contributions against py2/3 on Mac, win and Linux by travis-ci.com

t0b3:
fixes #98 and improves code quality checking across all platforms and python versions

PR #105, opened 2020-07-01
Inkscape v1.00 compatibility updates

uxwalter:
These changes get rid of all* deprecation warnings when running inkscape-silhouette on inkscape version 1.00.

I figured I send this out here for folks that were forced onto inkscape v1 and ran into issues with their silhouette cutter. Since the changes are likely not backwards compatible with 0.92 or even python2 I suspect you don't want to just integrate this. Let's either have a conversation about how to make this acceptable or use it as a simple guide to get the changes made properly by someone else.

It also fixes parts or all of issues #101 and #90, at least for me.

Disclaimers:
* Tested successfully on Inkscape v1.00, python3 and Silhouette Cameo 3. 
* Only tested with my typical workflows
* Didn't touch other files that probably need updating as well (e.g. silhouette_multi.py)
* Not tested on inkscape 0.92 and/or python2

### DIFF
--- a/silhouette/Graphtec.py
+++ b/silhouette/Graphtec.py
@@ -222,7 +222,7 @@ class SilhouetteCameo:
         try:
             for dev in usb.core.find(find_all=True):
               msg += "(%04x,%04x) " % (dev.idVendor, dev.idProduct)
-        except NameError: 
+        except NameError:
             msg += "unable to list devices on OS X"
         raise ValueError('No Graphtec Silhouette devices found.\nCheck USB and Power.\nDevices: '+msg)
 
@@ -409,7 +409,7 @@ Alternatively, you can add yourself to group 'lp' and logout/login.""" % (self.h
     elif isinstance(data, bytearray):
         return str(data).decode()
     else:
-        return data.tostring().decode()
+        return data.tobytes().decode()  # return data.tostring().decode()
 
   def try_read(s, size=64, timeout=1000):
     ret=None
@@ -690,7 +690,7 @@ Alternatively, you can add yourself to group 'lp' and logout/login.""" % (self.h
     s.enable_sw_clipping = sw_clipping
 
     # if enabled, rollers three times forward and back.
-    # needs a pressure of 19 or more, else nothing will happen 
+    # needs a pressure of 19 or more, else nothing will happen
     if trackenhancing is not None:
       if trackenhancing:
         s.write(b"FY0\x03")
@@ -958,7 +958,7 @@ Alternatively, you can add yourself to group 'lp' and logout/login.""" % (self.h
       #resp = s.read(timeout=40000) ## Allow 20s for reply...
       #if resp != "    0\x03":
         #raise ValueError("Couldn't find registration marks. (2)(%s)" % str(resp))
-      
+
       #resp = s.read(timeout=40000) ## Allow 20s for reply...
       #if resp != "    1\x03":
         #raise ValueError("Couldn't find registration marks. (3)")
@@ -975,7 +975,7 @@ Alternatively, you can add yourself to group 'lp' and logout/login.""" % (self.h
     #\xmin, ymin Zxmax,ymax, designate cutting area
 
     # needed only for the trackenhancing feature, defines the usable length, rollers three times forward and back.
-    # needs a pressure of 19 or more, else nothing will happen 
+    # needs a pressure of 19 or more, else nothing will happen
     #p = b"FU%d\x03" % (height)
     #p = b"FU%d,%d\x03" % (height,width) # optional
     #s.write(p)
@@ -1038,7 +1038,7 @@ Alternatively, you can add yourself to group 'lp' and logout/login.""" % (self.h
     for line in open(file,'r').readlines():
       if re.match(r'\s*\[', line):
         exec('data1234='+line)
-        break 
+        break
       elif re.match(r'\s*<\s*svg', line):
         print(line)
         print("Error: xml/svg file. Please load into inkscape. Use extensions -> export -> sendto silhouette, [x] dump to file")
--- a/sendto_silhouette.inx
+++ b/sendto_silhouette.inx
@@ -130,7 +130,7 @@ Always use the least amount of blade possible.
     <page name='about' _gui-text='About'>
       <param name="about_who" type="description">inkscape-silhouette extension from https://github.com/jnweiger/inkscape-silhouette by Jürgen Weigert [juergen@fabmail.org] and contributors</param>
       <!-- Keep in sync with sendto_silhouette.py line 78 __version__ = ... -->
-      <param name="about_version" type="description">Version 1.21</param>
+      <param name="about_version" type="description">Version 1.23</param>
     </page>
   </param>
 
--- a/sendto_silhouette.py
+++ b/sendto_silhouette.py
@@ -86,10 +86,11 @@
 # 2019-07-25, jw, v1.21 -- merge from github.com/olegdeezus/inkscape-silhouette
 #                          merge from fablabnbg
 # 2019-08-03, jw, v1.22 - added a copy of pyusb-1.0.2 as a fallback on any platform.
+# 2020-07-01, uw, v1.23 - compatability changes for inkscape version 1.00
 
 from __future__ import print_function
 
-__version__ = '1.22'	# Keep in sync with sendto_silhouette.inx ca line 79
+__version__ = '1.23'	# Keep in sync with sendto_silhouette.inx ca line 79
 __author__ = 'Juergen Weigert <juergen@fabmail.org> and contributors'
 
 import sys, os, shutil, time, logging, tempfile, math, re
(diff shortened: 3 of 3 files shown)


bianchifan:
System: Ubuntu 20.04, Inkscape 1.0, Python 3, Cameo 2

This should be plotted/cut
Bildschirmfoto von 2020-07-04 11-20-39
...and this was plotted, BLACK = regular plot, RED = with REGMARKS activated
DSC03436
error messages for regular plot
Bildschirmfoto von 2020-07-04 11-48-03
...and for REGMARKS plot
Bildschirmfoto von 2020-07-04 11-49-30

uxwalter:
Pretty sure those warnings come from the master version of Graphtec.py, not the one in this PR. Maybe double check that? While at it also check the version in the "About" tab although the absence of additional warnings suggests you got the correct version of sendto_silhouette.py (i.e., the one from this PR)

bianchifan:
> Pretty sure those warnings come from the master version of Graphtec.py
Yoe*re absolutely right..

> not the one in this PR. Maybe double check that?
What did I?
I followed the three commits you posted above.
I was wondering about no 2 and 3 because they point to the same change in "sendto_silhouette.inx".
I just made a compare regarding your forked master and found the other changes :)

Excuse me please that I didn't notice your fork :(

bianchifan:
Results of my tests with new Graphtec.py:
- deprecation warnings for regular plot are gone
- plot is still the same as above, some objects are missing

Dump with actual PR
Dump_Fox32_decals
To compare... dump with old extention and Inkscape 0.92.5 (Ubuntu 19.10)
Dump_Fox32_decals-0 92 5
OLD extention, because PR extention does not work with Inkscape 0.92.5
_Traceback (most recent call last):
  File "sendto_silhouette.py", line 125, in <module>
    from inkex.paths import CubicSuperPath, Path
ModuleNotFoundError: No module named 'inkex.paths'; 'inkex' is not a package_

- deprecation warnings for regmarks plot have changed
Bildschirmfoto von 2020-07-05 12-37-02

silhouette/Graphtec.py
```diff
@@ -409,7 +409,7 @@ def read(s, size=64, timeout=5000):
     elif isinstance(data, bytearray):
         return str(data).decode()
     else:
-        return data.tostring().decode()
+        return data.tobytes().decode()  # return data.tostring().decode()
```
t0b3, line 412: not compatible with python2:
AttributeError: 'array.array' object has no attribute 'tobytes'

thus another solution is needed to keep compatiblity with python2

uxwalter:
I have no means to check with python2 anymore, I could blindly add a "if PY3" around the one line change without verification.

I'd encourage you to read up on https://www.python.org/doc/sunset-python-2/ and consider changing direction to retain Python2 compatibility. Especially the part about security concerns.

t0b3:
@uxwalter I neither do. I personally would like to push forward.
Still I believe it's important to be aware of breaking changes, as I understood the maintainers still have use cases with Inkscape < 1.0.

To make sure any changes keep all the possible inkscape/python/os combinations running, I propose a PR to introduce continuous integration tests (#99).
image
NB: You may enable those tests by creating a travis-ci account and enabling travis on your github inkscape fork. This enables you to oversee the impact of code commits.

t0b3:
@EtherGraf would you still like to keep compatibility with python2 and Inkscape < 1?

If yes
* I propose to use 
`try: [...inkscape1_functions...]   #inkscape1.0`
`except: [...legacy_functions...]   #inkscape 0.95` 
patterns like these https://github.com/t0b3/inkscape-silhouette/blob/65aa3e2302701c741c56b219fd93dba0a7e65766/sendto_silhouette.py#L277-L373

If no 
* I presume it's ok to introduce some breaking changes like the ones proposed in this PR.

EtherGraf:
I agree, we need to be compatible with Inkscape 1.0 as soon as possible. 
My personal working environment is based on Xubuntu 18.08 and Inkscape 0.92.3 with PY2. So I think we should be compatible with PY2 at least until the extension is working fine with Inkscape 1.0

I see some possibilities to achieve this:
1. We do a fork and let the legacy PY2 stuff behind us in the old project.
2. We enclose the incompatible code in constructs like if PY2 ... else ... / try ... catch...
3. We introduce some compatibility layer where we hide the PY2/PY3 stuff.

I don't like the example you have given above, as it doubles the whole argument parser code. I think this is error prone for future development. It's not easy to oversee all such branches in future maintenance.

Is there no option parser which is supported in PY2 and PY3?
Would it be thinkable to write a wrapper function call, which delegates to the appropriate option parser?
Then the programm logic gets written down only one time.
e.g.

function addOption():
  if PY2 then:
    ... oldOptionParser.add
  else:
    ...newOptionParser.add

EtherGraf:
From my long running software projects I have a further tip:
In such cases I prefer to use the old version in the if condition:

e.g.
if PY2 then...
else...

and NOT 
if PY3...
else....

This has two advances. First I can extend the list
if PY2...
if PY3...
if PY4...

Second i can with small effort remove all legacy stuff: Simply remove all PY2 clauses.

t0b3:
I can confirm we are already compatible with Python 3 and Inkscape 0.92.5.
(my personal workspace) and thoroughly tested with Travis Continuous
Integration Tests.
To make this results public I suggest to merge PR #99 and create a free
Travis account for this project. This decision really won't do any harm.
Can you agree to this or would you like more detailed information?

I like your proposition to introduce
if PY2; if PY3
patterns. Unfortunately things will become more complicated as it's mainly
not PY3 but Inkscape 1.0 introducing breaking API changes. That is why I
went for patterns
try #inkscape 1.0...

OTOH it would be easier for maintenance to go for a separate branch without
all the legacy.


EtherGraf <notifications@github.com> schrieb am Sa., 18. Juli 2020, 10:55:

twmr:
I had to apply the following patch on top of this PR to fix an issue that I had with the extension.

```diff
diff --git a/sendto_silhouette.py b/sendto_silhouette.py
index 4050cf9..00d83ba 100644
--- a/sendto_silhouette.py
+++ b/sendto_silhouette.py
@@ -631,12 +631,12 @@ class SendtoSilhouette(inkex.Effect):
                                         if t:
                                                 newpath.set( 'transform', t )
                                         a = []
-                                        a.append( ['M ', [x, y]] )
-                                        a.append( [' l ', [w, 0]] )
-                                        a.append( [' l ', [0, h]] )
-                                        a.append( [' l ', [-w, 0]] )
-                                        a.append( [' Z', []] )
-                                        newpath.set( 'd', simplepath.formatPath( a ) )
+                                        a.append( ['M', [x, y]] )
+                                        a.append( ['l', [w, 0]] )
+                                        a.append( ['l', [0, h]] )
+                                        a.append( ['l', [-w, 0]] )
+                                        a.append( ['Z', []] )
+                                        newpath.set( 'd', str(Path( a ) ))
                                         self.plotPath( newpath, transform )
 
                         elif node.tag == inkex.addNS( 'line', 'svg' ) or node.tag == 'line':
```

tuxfunk:
I'm not sure if I'm applied the fixes correctly, but I was able to cut again with Inkscape 1.0.1 (not possible before with the official plugin). But I noticed a problem: I positioned a complex object (airbrush masks) in the upper right corner on the page. The object was correctly cut in size, but on the upper left corner of the page, so the position on the page was ignored somehow.
Thanks!

uxwalter:
I made an update a while ago that I figured I share here. The plot routine never applied any transform to the path being plotted which caused a number of issues. I lifted a few lines of code from gcodetools.py which is part of inkscape.

For starters, the awkward change to the unit to pixel math is now gone which makes a whole lot more sense. I assume that happened because my default unit is inches and the conversion from inches to mm never happened (but inches to pixels is 1-1). Just a guess but since it's working now without my previous change I won't fret about it.

@bianchifan : Probably more important is that your input file now renders correctly as far as I can tell. Give it a shot and let me know.

jnweiger:
I have not yet seen a nice exmples how to keep compatible with old py2 0.92.x inkscape at the same time as introducing 1.0 compatibility. 
I agree that readability suffers as we sprincle the code with if py2/ if py3, --- I have already done that pattern in other projects and I think it is acceptible. 
If nothing else helps, let us break compatibility with inkscape 0.92.x and move forward. 

@t0b3 I've invited you to join the team, 
@EtherGraf you had only writer, I made you maintainer now.

I'd love to see CI here too!

Sorry that I do so little myself, it really worries me that I can only create stuff and then apparently have to dump the maintenance burden on others. Sincerely sorry fo that bad behaviour on my side!

silhouette/Graphtec.py
t0b3: resolved in master:
`try: ...tobytes()... exept: ...tostring()...`

t0b3:
@jnweiger thank you for resending the invitation to the team. 
@uxwalter huge thanks for bringing up a working solution for Inkscape 1.0.

Finally I managed to get some time and to
- [x] make the code compatible to run with any Inkscape version. see latest code in master. with this code, choice is up to the people when to make the transition to Inkscape 1.0
- [x] make the basic tests run on any platform. see https://travis-ci.com/github/fablabnbg/inkscape-silhouette for more details.

You are very welcome to provide more tests. Also let me know your feedback.

t0b3:
@thisch @tuxfunk you are welcome to try with the latest code in master and let me know your feedback. 
as far as currently introduced tests tell the basic parts of the code work on any Inkscape version https://travis-ci.com/github/fablabnbg/inkscape-silhouette - but these are just the basic tests. Feel free to propose more tests ...

PR #107, opened 2020-07-09
Cameo4 support

FlyingSamson:
This PR offers:
* Support for Cameo 4:
   - all in all not much has changed in comparison to the Cameo 3
   - Silhouette Studio now also transmits acceleration to the device
* Extensive restructuring of the code (all credit for that great work goes to @mr-sven):
   - the new function `send_command` and `send_escape` now abstracts away all those control characters (0x03, 0x1b, etc.)
   - the new class `SilhouetteCameoTool` and multiple helper functions abstract away the command strings themselves and therewith allows for a more tidy `setup` and `plot` functions where self-explanatory names show directly which machine parameter is changed
   - Conversion between SU and mm is now done through a function which also takes care of rounding to the nearest `int`, making the code more self-explanatory and clean.
   - Cleanup of doc strings


t0b3:
@FlyingSamson Thank you very much for contributing the Cameo4 instructions and the function abstractions.
Unfortunately the automated tests fail (see https://travis-ci.com/github/fablabnbg/inkscape-silhouette/builds/211303354) and my Cameo3 stopped working.

Can you provide fixes to get the code running? I'd like to merge your PR as soon as the tests succeed and my Cameo3 still works.

FlyingSamson:
@t0b3 I think I fixed the issue with the CI pipeline. The function handling the "save" send was not able to handle empty command lists, which I presume the umockdev test was creating by only connecting to the device without sending any plot commands, correct?

Can you test again on your Cameo 3 and maybe elaborate a little further on what is going wrong?

t0b3:
@FlyingSamson CI tests pass now. However dumps take a really long time now.
`python sendto_silhouette.py --dummy=True examples/testcut_square_triangle.svg` takes now 28 instead of 0 sec while counting to 15.
Can you please fix the behaviour to dump instantly again?

t0b3:
@FlyingSamson tracked down the unexpected long duration to [this changed line](https://github.com/FlyingSamson/inkscape-silhouette/blame/1f29c2a9f0669076d26fb28a13b44a201896b520/silhouette/Graphtec.py#L1286) 
- what happens on your Cameo4 if we omit this line calling the `wait_for_ready()` method?
- if you think it is absolutely necessary, please propose a way to have `--dummy=True` dumps work speedy again...

silhouette/Graphtec.py
```diff
@@ -973,37 +1249,41 @@ def plot(s, mediawidth=210.0, mediaheight=297.0, margintop=None,
...
+        self.draw_mm_cmd(bbox['ury'], bbox['urx']),
+        self.draw_mm_cmd(bbox['lly'], bbox['urx']),
+        self.draw_mm_cmd(bbox['lly'], bbox['llx']),
+        self.draw_mm_cmd(bbox['ury'], bbox['llx'])]
+
     # potentially long command string needs extra care
-    s.safe_write(p)
+    self.safe_send_command(cmd_list)
+
+    self.wait_for_ready()
```
t0b3: @FlyingSamson can we omit this line?

t0b3: let's call `wait_for_ready()` only for real devices but not for dumps

t0b3:
tested here with Cameo3 (Linux, python3.9, Inkscape101) - works fine
@jnweiger @EtherGraf do you prefer to test these changes with your own plotters before we merge them?

EtherGraf:
I have seen, that there have been a lot of changes the past days. 
If I find some time next days, then I will fetch my equipment and will do a regression test with my old inkscape 0.92/PY2 installation and my Cameo 3. I cannot guarantee it, but I will try.

t0b3:
@EtherGraf I think the most important thing to test would be this PR #107 with just your current equipment. This would already be great.

FlyingSamson:
> @FlyingSamson CI tests pass now. However dumps take a really long time now.
> `python sendto_silhouette.py --dummy=True examples/testcut_square_triangle.svg` takes now 28 instead of 0 sec while counting to 15.
> Can you please fix the behaviour to dump instantly again?

I still could not figure out why this would only happen after the changes made in this PR. I mean the call to `wait_for_ready()` was present in the original code as well, and the implementation of `wait_for_ready()` wasn't altered either. It might be working now, but it makes me wonder whether there could be something else behind it.

FlyingSamson:
> @FlyingSamson tracked down the unexpected long duration to [this changed line](https://github.com/FlyingSamson/inkscape-silhouette/blame/1f29c2a9f0669076d26fb28a13b44a201896b520/silhouette/Graphtec.py#L1286)
> 
> * what happens on your Cameo4 if we omit this line calling the `wait_for_ready()` method?
> * if you think it is absolutely necessary, please propose a way to have `--dummy=True` dumps work speedy again...

I will perform another test with my machine in the next days. I can't really remember why I put this call there in the first place. From looking at the code it seems unnecessary, since `safe_write()` which is called by `safe_send_command()` will wait for `ready` on the last chunk.

t0b3:
> > @FlyingSamson tracked down the unexpected long duration to [this changed line](https://github.com/FlyingSamson/inkscape-silhouette/blame/1f29c2a9f0669076d26fb28a13b44a201896b520/silhouette/Graphtec.py#L1286)
> > 
> > * what happens on your Cameo4 if we omit this line calling the `wait_for_ready()` method?
> > * if you think it is absolutely necessary, please propose a way to have `--dummy=True` dumps work speedy again...
> 
> I will perform another test with my machine in the next days. I can't really remember why I put this call there in the first place. From looking at the code it seems unnecessary, since `safe_write()` which is called by `safe_send_command()` will wait for `ready` on the last chunk.

@FlyingSamson never mind, I introduced a fix in your branch: see 84975e9

EtherGraf:
@t0b3 Can you give me a quick hint, how to checkout this PR107 branch from github to my local repo?

t0b3:
> @t0b3 Can you give me a quick hint, how to checkout this PR107 branch from github to my local repo?

`git fetch origin pull/107/head:pr/107 && git checkout pr/107`
see https://github.community/t/checkout-a-branch-from-a-fork/276

EtherGraf:
Thanks for your valuable work and the tip. 
I have tested the extension with my Ubuntu 18.04 linux, Inkscape 0.92 with python2. My device is a Silhouette Portrait.
Just done a small testcut without any specialities like regmarks and not checked dimension for exactness.

At first I have tested the master branch. Worked without any flaws. Thumbs up!

Then I tested the PR107 branch. 
On first glance worked too without problems. But after succesfully cutting Inkscape shows a message about additional data returned from the script, containing only a single empty line. Looks like an empty print statement to stdout. Testing previous versions it looks like the fault has been introduced in caa59f3098ea8c9b5c58a91fbea1ab654c22f913 and removed already in master. 

If I merge this #PR107 into master, it does NOT work anymore. 
Just get a error message " 1/15: status=moving", but device does not move.

Needs more investigation.

EtherGraf:
I cannot reproduce the faults any more.
There was some hickup of my device, causing cranky noises at each start, which disappeared only after a power cycle reset. Maybe it gets this hickup by some unknown Cameo4 commands with an intermediate version?

At the moment PR107 looks working to me. 

The only nuisance is the error box with a blank line at the end of the plot. This was found to be introduced in the big 72f5e882655a88511c3340503d42085bd3905994. No idea, which line is the faulty one.

EtherGraf:
The error box is caused by the call to wait_for_ready() in graphtec.py:
Inside this function is a line 

`if verbose: print("",file=sys.stderr)`

If I comment this line out, the error box at print end does not appear.

t0b3:
> If I merge this #PR107 into master, it does NOT work anymore.
> Just get a error message " 1/15: status=moving", but device does not move.
> 
> Needs more investigation.

In case such hicups reappear, one solution would be to omit these two lines:
https://github.com/FlyingSamson/inkscape-silhouette/blob/84975e9a19b406e3982f8cfedd60f59678776f82/silhouette/Graphtec.py#L1286-L1287

t0b3:
> The error box is caused by the call to wait_for_ready() in graphtec.py:
> Inside this function is a line
> 
> `if verbose: print("",file=sys.stderr)`
> 
> If I comment this line out, the error box at print end does not appear.

@FlyingSamson can you delete this line in your PR?

FlyingSamson:
I just tested without those lines on my Cameo 4 and it seems to work just fine. So I will remove those lines for now.

FlyingSamson:
Is the branch `prepare_inkscape10port` that you included in the PR supposed to be free of deprecation warnings? Because I still get one:

```
/Users/dominik/Library/Python/2.7/lib/python/site-packages/usb1/__init__.py:2208: DeprecationWarning: Use "with USBContext() as context:" for safer cleanup on interpreter shutdown. See also USBContext.open().
  DeprecationWarning,
```
Current setup is 
- Inkscape 1.0 (4035a4f, 2020-05-01)
- Python 2.7.16
- Mac OS X 10.13.6

Not sure how to fixed that atm, as we clearly can't wrap everything in a `with ... as ...` statement.

t0b3:
@FlyingSamson I didn't expect Inkscape to run with Python2 version at all. I'd rather upgrade some recent Python3 version. For OSX I can just tell you this warning also appears in [automated tests](https://travis-ci.com/github/fablabnbg/inkscape-silhouette/jobs/467490618). 
Any help to avoid this warning will be welcome... though I guess this needs to be fixed [upstream](https://github.com/vpelletier/python-libusb1)

PR #112, opened 2020-09-27
Fix str object has no decode() method error. Follow-up on #94

FlyingSamson:
On OSX the bulkRead method is used. That returns a bytearray. The elif branch gets triggered, which, when using Python3 (and Inkscape 0.92.5), fails as the constructed str object (by calling str(data)) does not posses a decode() method in Python3 any more. I don't saw any reason why we should not handle the bytearray case similarly to the str and bytes case.

Please correct me if I am wrong and we indeed need to handle this case separately.


t0b3:
@FlyingSamson thanks for pointing out [the bytearray case](https://github.com/fablabnbg/inkscape-silhouette/blob/fc4dc2a631cc60d81a00e9ecf9179c90b9c101ae/silhouette/Graphtec.py#L409) is triggered on OSX and the handling should be the same for *(str,byes,bytearray)*

PR #120, opened 2021-01-02
Add cutting mat templates

FlyingSamson:
This PR provides two templates showing the 12" x 12" and the 12" x 24" mats on a background layer in Inkscape. I found those useful for visual guidance during positioning of objects for cutting. Maybe other plugin users find them helpful too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,13 @@ sudo cp -R silhouette /usr/share/inkscape/extensions/
   * **Pressure**            Custom Pressure on the blade. One unit is said to be 7g force.
 7. Press Apply button to start cut.
 
+## Templates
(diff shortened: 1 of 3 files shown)


twmr:
@FlyingSamson Thx a lot for adding two templates for the mats. I was looking for svg files of the mat of my Silhouette Portrait, but couldn't find one until I found your PR. Unforunately your templates are only for 12''x12'' and for 12''x24'' mats, mine however is a 8''x12'' mat. Could you please also add a template for this mat (I'm not yet so familiar with inkscape to adapt your template to the new geometry)? Thx!

t0b3:
see #164 for the very much needed information on 8x12

Issue #126, opened 2021-01-31
Incorrect extensions folder for Mac


jtebert:
I'm installing the extension on Mac OS 11.1 (Big Sur) and Inkscape 1.0.1, and this has the incorrect extensions path hard-coded into the installer script. (I haven't tried this on an older version of Mac to know if this is a new problem or carries over from older versions of Mac OS.)

In `install_osx.py`, the `extensions_dir` is set as:
```python
extensions_dir = os.path.join(os.path.expanduser("~"), ".config", "inkscape", "extensions")
```

However, within the Inkscape (1.0.1) preferences, the extension directory is listed as:
```
/Applications/Inkscape.app/Contents/Resources/share/inkscape/extensions
```

As a result, the install script does not produce an error, but (of course) the extension doesn't appear in Inkscape. If I change the extensions directory to the one listed in Inkscape, I'm able to install it.

Is there a way for the extension to pull the directory from Inkscape more directly so that it will have the correct location?

t0b3:
I'd be interesting to know what Inkscape 1.0+ on OSX tells about user defined extensions...
what path is shown at menu Edit > Preferences > System > User extensions? - see https://wiki.inkscape.org/wiki/index.php?title=File:Open_folders_from_preferences.png

Further I'm curious to see if the extension would be correctly recognized if you place at _user extensions_ path...

tschamm:
I installed the extension on Mac OS 11.2 (Big Sur) with Inkscape 1.0.2, for me the correct Install Folder on OS 11 is:
```
~/Library/Application Support/org.inkscape.Inkscape/config/inkscape/extensions
```
This is also the path given, when opening the Inkscape Settings -> System -> User settings.

@jtebert Can you check if this is valid for you too?

PR #127, opened 2021-02-06
Update install instructions and install script for Mac OS.

### LINKED ISSUE #81: no module name usb 1
I got it it installed on a Mac 10.13.6 

the machine is connected with the usb and usb.core.find() gives me thins

<DEVICE ID 0b4d:112b on Bus 020 Address 007>

and usb.version_info[0] gives me 1

I get this error when I run the plugin in inkscape .. 

thanks for any help

Screenshot 2019-03-25 21 59 57

tschamm:
Update install instructions and install script for Mac OS.
* Install python3 via brew
* Use pip3 to install dependencies (easy_install is deprecated)
* Update user extension dir.
* Fixes #81 and #126

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,7 +95,7 @@ sudo cp -R silhouette /usr/share/inkscape/extensions/
   * `brew install python3`
 * Install the extension:
   * `./install_osx.py`
-  * Add brew python for user extensions (see [Extensions_Interpreters](https://wiki.inkscape.org/wiki/index.php/Extension_Interpreters)):
+  * Add brew python for user extensions. For details on selecting a specific interpreter version see [Inkscape Wiki - Extension Interpreters](https://wiki.inkscape.org/wiki/index.php/Extension_Interpreters#Selecting_a_specific_interpreter_version_.28via_preferences_file.29):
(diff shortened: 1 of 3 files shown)


tschamm:
I'm aware this PR would drop python2 support on Mac OS. 

If python2 support is still needed on Mac OS, we have to look for a solution on how to find a compatible version of python2 libusb package and the libusb dylib installed by `homebrew`.

t0b3:
@jtebert can you try if this proposed solution works for your setup?

jtebert:
Thanks for making these updates.

There seems to be just one change necessary for this to install correctly. When I try to install and run it as is, it complains that numpy is not installed. If I install numpy it appears to work. (I'm not with the cutter to confirm that it cuts correctly, but the only error I now get is that it can't find the plotter.)

I believe that adding numpy to the requirements.txt should fix this.

One other small suggestion -- it might be useful to give more detailed instructions in-line in the readme about how to update the Inkscape python interpreter.

tschamm:
Any update on this? The comments from @jtebert were addressed by the commits done on Feb, 9th.

t0b3:
@tschamm thanks for this contribution

PR #136, opened 2021-06-04
feat: Add Cameo 4 Pro cutter and stubs for Cameo 4 Plus

gwhitney:
Adds the proper USB ID and parameters for my Cameo 4 Pro cutter. I included my best guess for what the values should be for the Cameo 4 Plus but left them commented out since I do not have one to verify. But I created aliases for the whole Cameo 4 product line in the code, so that if the Cameo 4 Plus code is uncommented (and possibly corrected as need be), all the rest of the code should work without further modification.

This change is independent from my recent PR #135 but I absolutely used that one in developing this, so to me that's an argument  for merging that one as well.

### DIFF
--- a/silhouette/Graphtec.py
+++ b/silhouette/Graphtec.py
@@ -19,6 +19,7 @@
 # 2016-05-21  detect python-usb < 1.0 and give instructions.
 # 2017-04-20  Adding Cameo3 USB IDs
 # 2020-06-    Adding Cameo4 and refactor code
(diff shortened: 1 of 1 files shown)


t0b3 (changes requested):
date 2020-06-03 is one year off

gwhitney:
Oops, sorry missed a year due to pandemic ;-) Fixed now.

PR #137, opened 2021-06-04
feat: Add a floating point tolerance to the software path clipping

gwhitney:
Once I got inkscape_silhouette talking to my Cameo 4 Pro, I found it was often not cutting lines that touched the boundary of my design. I tracked this down to what appear to be floating point rounding issues in the unit conversions. (The bounding box was coming out to 48.0 mm but some points on its boundary were getting coordinates like 48.000056.) So this PR adds a settable tolerance when comparing points for clipping (points within the tolerance of the boundary will not be clipped). It also reduces the repeated code in the software clipping for the first and subsequent points of a path.

This PR is completely independent of #135 and #136.

### DIFF
--- a/silhouette/Graphtec.py
+++ b/silhouette/Graphtec.py
@@ -393,6 +393,7 @@ Alternatively, you can add yourself to group 'lp' and logout/login.""" % (self.h
     if self.dev is None or 'width_mm' in self.hardware:
       self.leftaligned = True
     self.enable_sw_clipping = True
+    self.clip_fuzz = 0.05
 
   def product_id(self):
     return self.hardware['product_id'] if 'product_id' in self.hardware else None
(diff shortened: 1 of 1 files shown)


EtherGraf:
I think, this setting should not be user settable. This is an internal rounding error and the solution should be hardcoded inside the code. There is no use for the user in changing this setting.

In optimal case the rounding should be done by conversion to integer device units. Than the clipping comparison could be executed with the integer values. This would be the exact solution. But I assume this will need some more refactoring of the code. 

So as an alternative I would suggest setting the clip fuzzing to an constant of 1/20mm, as this is the device resolution. 

Removing the duplicated code is an good thing, I had the same idea a year ago, but never finished my patches.

t0b3 (changes requested):
can you adapt the PR as @EtherGraf proposes?

gwhitney:
Done; I eliminated the user-settable argument but otherwise left the clip_fuzz in (only affects silhouette/Graphtec.py now), with a default of 0.05 mm.

gwhitney:
Hmm, I see the test failures. Maybe setup is not called in the test. I will modify to ensure clip_fuzz is initialized and push the change.

silhouette/Graphtec.py
```diff
@@ -1270,6 +1266,7 @@ def plot(self, mediawidth=210.0, mediaheight=297.0, margintop=None,
     bbox['clip'] = {'urx':width, 'ury':top, 'llx':left, 'lly':height}
     bbox['only'] = bboxonly
     cmd_list = self.plot_cmds(pathlist,bbox,offset[0],offset[1])
+    print(f"Final bounding box and point counts: {bbox}", file=self.log)
```
t0b3: @gwhitney CI tests are failing probably due to this line. What happens if you drop it?

t0b3 (changes requested):
CI tests are failing....

gwhitney:
Oh, I thought it was something else, I just pushed a fix. Will check again in a moment.

But uh, wait, does the python source code have to run under _both_ Python 2 and 3? I thought 2 was now considered "end-of-lifed". I am pretty sure that Python 2 did not have f"" strings. If this needs to support Python 2, let me know and I will figure out how to change that log message to not use an f"" string.

t0b3 (changes requested):
legacy python2.7 is still failing - can you take care of it and replace the f"" string
though it's EOL it might still be used in some fablabs legacy setups

silhouette/Graphtec.py
gwhitney: Modified to use str() which is valid in both Python 2 and 3

gwhitney:
Yup, f"" now gone, sorry I am living in a blissful Python3-only world ;-)

PR #138, opened 2021-06-05
(Do not merge) First pass at enhancing logging and dumping features

gwhitney:
This PR is inspired by and intends to replace #69. Many thanks to @neocogent for the model. I am doing a large production run of identical pieces on my Cameo 4 Pro and so need the commands in a file to be sent repeatedly. The main differences from #69 are as follows:

- Based on current "master" branch (By the way, please would you consider renaming this branch to "main" as part of the effort to make tech language more inclusive?)
- Makes whether the commands are captured independent of whether they are being sent to the cutter (so you can capture the commands on a "live" run as well)
- Still queries for a device when dumping, as the identity of the device can affect the commands produced.
- Allows you to toggle whether the queries to the device are included in the command capture, or just the directives.

In addition, this PR allows both a dump of the raw paths and a capture of the cutter commands on the same run. Since the raw paths are primarily for debugging, once there were files for the log and the commands, it seemed excessive to have a third file for the paths, so I moved those to the log, which simplified the code in that a separate tempfile did not then have to be generated for them. (If it's important that the paths dump goes into a separate file, just let me know and I can revert that bit of the changes.)

Finally, because there are now several options related to logging and dumping, this PR moves those controls to their own page in the GUI.

### DIFF
--- a/sendto_silhouette.inx
+++ b/sendto_silhouette.inx
@@ -104,9 +104,13 @@
       <param name="mintravel_help" type="description">Minimal Traveling: Find the nearest startpoint to minimize travel movements</param>
       <param name="mintravelfull_help" type="description">Minimal Traveling (fully optimized): Additionally search startpoints in closed paths</param>
       <param name="sw_clipping" type="boolean" _gui-text="Enable Software Clipping">true</param>
-      <param name="dummy"    type="boolean" _gui-text="Dummy device: Send to /tmp/silhouette.dump">false</param>
-      <param name="dummy_help" type="description">Used for debugging and developent only!</param>
+    </page>
+    <page name='logdump' _gui-text='Log and Dump'>
       <param name="logfile" type="string" _gui-text="Save log messages in file:"></param>
+      <param name="log_paths" type="boolean" _gui-text="Include final cut paths in log (for debugging)">false</param>
+      <param name="cmdfile" type="string" _gui-text="Transcribe cutter commands to file:"></param>
+      <param name="inc_queries" type="boolean" _gui-text="Include cutter queries in command transcript">false</param>
+      <param name="dry_run" type="boolean" _gui-text="Dry Run: do not send commands to device">false</param>
     </page>
 
     <page name='blade' _gui-text='Blade Setting'>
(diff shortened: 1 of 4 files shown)


gwhitney:
Oops, I had eliminated tempfile from the main code, didn't realize it was used in the testing.

gwhitney:
Oops no. 2, tests now need to use --dry_run option instead of --dummy (retired the latter since it is a combo of --dry_run and --log_paths).

gwhitney:
Oops no. 3, the log does not go to standard output in the tests as I had assumed it did. So bagging the check of the output about it being a dummy run altogether; not clear what that assertion was accomplishing.

gwhitney:
OK, with all of those oopses fixed, almost all of the tests are passing. Only the Python 2 umockdev cameo3 test is failing. However I don't have much of a clue as to why, especially given that the Python 3 umockdev test succeeds. I confess I do not really understand what is going on with the umockdev tests. If you can give any pointers as to how to debug this, especially in my Python 3-only world, I would be very grateful and happy to make whatever effort is needed to get this ironed out quickly so that it's possible to generate command files, which is critical for my operations. Thanks!

t0b3:
This contribution looks promising, though there are some things to look at.
Instead of one complex commit we generally prefer multiple single commits, i.e. one per chore/change, in order to reduce complexity and simplify the code review.

Can you split it to multiple commits, i.e.
1. dummy -> dry_run
2. combine dump & log (beware dump format and `misc/read_dump.py`)
3. ...

This will help a lot. Thank you.

gwhitney:
OK, will do. Should I leave it as a single PR once it is split up, or divide it into more than one PR? Thanks for the guidance.

t0b3:
In this case multiple PR would be helpful to track down the CI test results. Thanks for your contribution. Appreciate.

gwhitney:
OK I am pursuing that strategy and so marked this as (Do not merge). I am happy to have it stay open until I have all of the pieces submitted as independent PRs, but if you prefer to close it before then I understand.

gwhitney:
All functionality incorporated via #140, #142, and #143. Closing.

PR #139, opened 2021-06-05
feat: Greater control over orientations of paths

gwhitney:
This PR adds an option to pre-orient paths to be monotone in either the Y or X direction. The motivating use case here is that I am using the Cameo to score without a mat, and this only works if the material is pulled through the tool (descending Y coordinate); it is impossible to _push_ a flexible, unbacked material into a scoring tool.

It also adds a variant of the MinTravel strategy which declines to reverse the orientation of any paths it is given. That variant of the strategy is more useful when path orientations are critical.

As before, this PR is totally independent of #138, but I did make use of the features of #138 in developing this.

### DIFF
--- a/sendto_silhouette.py
+++ b/sendto_silhouette.py
@@ -94,7 +94,7 @@ from __future__ import print_function
 __version__ = "1.24"     # Keep in sync with sendto_silhouette.inx ca line 79
 __author__ = "Juergen Weigert <juergen@fabmail.org> and contributors"
 
-import sys, os, time, tempfile, math, re
+import sys, os, time, tempfile, math, operator, re
 
 
 # we sys.path.append() the directory where this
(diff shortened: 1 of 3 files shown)


sendto_silhouette.py
```diff
@@ -1144,6 +1144,8 @@ def write_progress(done, total, msg):
                         if len(newpath) == 1:
                             # Have to make some progress:
                             newpath = [curpath[-1], newpath[0]]
+                            # Don't leave behind an orphan
+                            if len(curpath) == 1: curpath = []
```
t0b3: newline missing?

t0b3 (changes requested):
looks good otherwise
did you test it on your Cameo?

sendto_silhouette.py
gwhitney: Oh, is this better?

gwhitney:
And yes, I am now producing nice scores with the "desy" pre-orientation and the MinTravel (no reversal) strategy. :)

PR #140, opened 2021-06-05
Move cut path dump into log file and separate "dry run" behavior from "path dump"

gwhitney:
This PR is intended ultimately to replace #139. I have created it to divide the commit that made that PR into smaller self-contained commits as requested in the feedback. I am taking the liberty of submitting those commits one at a time through this PR because I am not sure how else to run the test suite. Thanks for your understanding, and if there's a different process you prefer I use please just let me know. I will remove the (WIP) designation and close #139 once all of the functionality there is incorporated here.

### DIFF
--- a/misc/read_dump.py
+++ b/misc/read_dump.py
@@ -1,15 +1,52 @@
-#!/usr/bin/python
+#!/usr/bin/env python3
 # this script reads inkscape dumpfiles and shows the plotter path
 import sys
 import matplotlib.pyplot as plt
-filename = sys.argv[1] if len(sys.argv)==2 else "/tmp/silhouette.dump"
-f = open(filename)
-lines = f.readlines()
-f.close()
-dump = eval(lines[-1])
+from pathlib import Path
+
+# The below is not correct under Windows; please help to correct.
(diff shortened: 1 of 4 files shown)


t0b3:
I guess you meant #138 instead, right?

gwhitney:
Yup. And per your comment in #138, I will finish up this PR with one more commit, to move the dump into the log and modify misc/read_dump.py to handle logs as well as old dumps.

t0b3:
OK, just tell me when you're done with it...

gwhitney:
OK, I think this portion is ready. Performing a dry run and dumping the cut paths are now separately selectable behaviors, so you can dump the cut paths from a live run if you want. The cut paths go into the log file, and misc/read_dump.py is updated so that it will handle log files (and in fact try to find the log file from inkscape's preferences, at least on Linux). Let me know what you think and then I will proceed from here.

gwhitney:
Oh, I also added some faint directional arrows and a start (green) and end (red) indicator to the display in misc/read_dump.py -- I wish I had known about read_dump when I was working on the orientations, it would have made life much easier!

PR #141, opened 2021-06-05
Graphical cut preview

### LINKED ISSUE #73: optimized order for rolls
I was cutting names in 2 columns down 4' of vinyl and the only way I could figure out how to do it was to select 6" of letters then cut, select the next 6" and cut, etc etc. Otherwise, it wants to cut one column all the way from 0 - 4' then go back to the top of the vinyl and start on the 2nd column.

The names were originally done in 2 rows and then rotated 90 degrees and the page resized to the object.

Otherwise, really great software.  KUbuntu 18.04.1 on a Silhouette Portrait 2

gwhitney:
Note this PR is on top of #140, and should only be merged if/after that is merged. It is just a small tweak to the revised read_dump.py (now moved into silhouette/ so that it would be installed with the extension) that provides a selectable preview before cutting. This PR should be a fast-forward once #140 is merged.

Now, it might be nice if that preview had "Cut" and "Cancel" buttons on it. That is definitely doable with matplotlib, and I am volunteering to add that feature if you want it in inkscape_silhouette. Just let me know, and let me know if you are thinking of merging this PR as it stands, or would like to wait and have me incorporate the buttons into this PR, or whether you prefer not to have a graphical preview as part of the inkscape extension interface at all (i.e. will be rejecting this PR). Thanks!


gwhitney:
Hmmm, with this change the inkscape_silhouette extension would have to formally depend on matplotlib, which is not strictly necessary without this change. Do you want to add that dependency and update the testing environment (and if there is anywhere in the source the dependency would have to be recorded, let me know)? Or do you want to make it an optional dependency for those who wish to see previews (and if so, any thoughts on how to make matplotlib only be imported if the preview option is specified? I am not familiar with conditional importing in Python)? Or do you want to abandon previewing as a mainstream feature in the inkscape extension?  Thanks for letting me know.

gwhitney:
Ah, I looked into it a bit; to make this an optional dependency where you can preview if you have matplotlib, I can just wrap the import in a try block and set a flag if it fails. So that's definitely one possible route here. Just let me know which way you want to go.

t0b3:
Great contribution, I'd like to integrate with an optional dependency on `matplotlib` and some simple GO / Abort buttons.
This parameter will need some hint / error msg for the matplotlib dependancy.

gwhitney:
On it; will rebase this on master and make the requested changes, then remove the (WIP) designation and let you know.

gwhitney:
OK, now it won't error if matplotlib is not present, but there are hints that you should install it; and it has Cut and Cancel buttons when used as a preview. (You can still run silhouette/read_dump.py exactly as misc/read_dump.py before.) Seems to be working well for me. Let me know if there's any other tweaks you'd like me to make, or feel free to merge.

t0b3:
well done!

PR #142, opened 2021-06-06
Allow device queries in dry_run mode

gwhitney:
Continuing the series of PRs to build up the functionality of #138. This one is primarily preparation for the transcription of command files. However, I've split it out as a separate PR because this is the change that entails the most significant low-level alterations: all of the low-level read/writes must now operate without a device, mocking up a perpetual 'ready' state and returning a dummy string when the device firmware version is queried. To facilitate and clarify this operation, the low-level routines now operate specifically with byte strings, and just the higher-level routines (like send_receive_command) deal in ordinary Python strings.

Hence, it seemed as though whatever issue was causing the Python 2 umockdev failure in #138, if it is going to recur with this restructuring of the changes, would be most likely to occur in this PR.

This PR is built on #140 and should be a fast-forward if/when that is merged.

### DIFF
--- a/sendto_silhouette.py
+++ b/sendto_silhouette.py
@@ -332,7 +332,7 @@ class SendtoSilhouette(inkex.Effect):
                 help="Include final cut paths in log")
         self.arg_parser.add_argument("--dry_run",
                 dest = "dry_run", type = inkex.Boolean, default = False,
-                help="Do not communicate with device")
+                help="Do not send commands to device (queries allowed)")
         self.arg_parser.add_argument("-g", "--strategy",
                 dest = "strategy", default = "mintravel",
                 choices=("mintravel", "mintravelfull", "mintravelfwd", "matfree", "zorder"),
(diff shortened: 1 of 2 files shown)


gwhitney:
Great, the Python 2 umockdev test is passing. I did this as a rewrite, not cutting/pasting changes from the first pass, so perhaps there was a subtle error or typo last time around that is OK this time. In any case, I will go on to doing the capture of cutter commands on top of this.

gwhitney:
Hmm, but I see a MacOS/xcode error in the Travis build. There's not really much information there, so not sure how I can debug. Please let me know how I can assist.

t0b3:
can you rebase onto master / resolve the conflicts? 
NB: let's see if travis still refuses on MacOS...

gwhitney:
OK, needed to beef up the exception handling a bit for Windows, but all looks well now.

t0b3:
well done

PR #143, opened 2021-06-06
feat: Option to save all cutter commands to file.

gwhitney:
Culmination of dividing #138 into three separate PRs. This is built on top of #142 and adds the real target feature: saving of cutter commands to a file. This can be done in a dry run and sent to the cutter later, or captured from a live run to be re-sent to make copies.

Ah, should I add documentation somewhere that on linux at least, saved commands can be later executed by the cutter by sending them to the appropriate /dev/usb/lpX file? If so, where should that information go?

For me, this feature will greatly enhance the usefulness of inkscape_silhouette, since I often make numerous copies of the same cut. Thank you for working with me on it.

### DIFF
--- a/sendto_silhouette.inx
+++ b/sendto_silhouette.inx
@@ -114,9 +114,14 @@
       </param>
       <param name="orient_help1" type="description">Note: Some strategies like "Without Mat" may reverse some path orientations, so final cut may not strictly obey orientation chosen above.</param>
       <param name="sw_clipping" type="boolean" _gui-text="Enable Software Clipping">true</param>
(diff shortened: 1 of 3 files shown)


gwhitney:
OK updated to be on top of the latest update to the primary branch, tests are looking good.

t0b3:
great contribution!
wouldn't it be worth to add a test case? 
i.e. `sendto_silhouette.py --dry_run=True --cmdfile cmdfile.out testcut_square_triangle_o.svg` and assert cmdfile.out contents match reference.out

gwhitney:
Thanks t0b3 for adding the test case for me, very much appreciated.

PR #144, opened 2021-06-07
fix: Don't worry about machine being ready in dry run

gwhitney:
With the latest changes, found an unfortunate state: if you were doing a dry run with the device connected but the machine happened to be in an unloaded state, inkscape_silhouette might silently wait forever for you to load media. So this PR unifies all waiting for ready through the (existing) wait_for_ready method, and then short-circuits that waiting in case you're in a dry run.

### DIFF
--- a/silhouette/Graphtec.py
+++ b/silhouette/Graphtec.py
@@ -586,9 +586,7 @@ Alternatively, you can add yourself to group 'lp' and logout/login.""" % (self.h
       # strip string candidate of unfinished command at its end
       safechunk = candidate[0:(candidate.rfind(CMD_ETX) + 1)]
       self.write(data = safechunk, is_query = False)
-      # wait for cutter to finish current chunk, otherwise blocking might occur
-      while not self.status() == "ready":
-        time.sleep(0.05)
+      self.wait_for_ready(timeout=120, poll_interval=0.05)
       so += len(safechunk)
 
   def send_command(self, cmd, is_query = False, timeout=10000):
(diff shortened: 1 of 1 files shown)


silhouette/Graphtec.py
```diff
@@ -686,21 +684,25 @@ def get_tool_setup(self):
...
         raise NotImplementedError("Waiting for ready but no device exists.")
-      if verbose: print(" %d/%d: status=%s\r" % (i, int(timeout*.5), state), end='', file=sys.stderr)
+      if verbose: print(" %d/%d: status=%s\r" % (i, npolls, state), end='', file=sys.stderr)
       if verbose == False:
         if state == 'unloaded':
-          print(" %d/%d: please load media ...\r" % (i, int(timeout*.5)), end='', file=sys.stderr)
-        elif i > 5:
+          print(" %d/%d: please load media ...\r" % (i, npolls, state), end='', file=sys.stderr)
+        elif i > npolls/3:
           print(" %d/%d: status=%s\r" % (i, int(timeout*.5), state), end='', file=sys.stderr)
```
t0b3: did you intentionally leave 702/704 untouched or would you want to change?
`-            print(" %d/%d: status=%s\r" % (i, int(timeout*.5), state), end='', file=sys.stderr) `
`+            print(" %d/%d: status=%s\r" % (i, npolls, state), end='', file=sys.stderr) `

t0b3 (changes requested):

silhouette/Graphtec.py
gwhitney: Great catch, thanks, dunno how I missed that one, sorry. Should be fixed now.

PR #146, opened 2021-06-08
feat: Allow model of cutting device to be overridden

gwhitney:
This is the last element (I know of) to fully allow @neocogent's use case to be handled by inkscape_silhouette (and it makes my life easier, too). If you want to generate a cutter command file without the cutting device attached to your computer, you need a way to set what model the device is (since the commands generated depend on the model). This PR allows you to override the detected model of the device (which is just a dummy when there is no device) with any of the known models. It also proceeds with "device initialization" even if there is no device. With these changes, it is possible to produce full, working command files for a Silhouette cutter on a computer that has no connection to any cutter, and then send those command files to an appropriate cutter attached to a different computer.

In case it's any relief, I am not planning any further PRs at the moment, unless a currently unknown issue comes up.


gwhitney:
Oops, allowing the initialization to occur without a device changes the default command file produced by inkscape_silhouette (as does running it when I have my Cameo 4 Pro plugged in!) so had to update the reference command file.

PR #149, opened 2021-06-13
Py39 fixes

fcharlier:
Hi,

Trying a newly acquired secondhand Silhouette Portrait (1st gen) on OSX, inkscape-silhouette first crashed with `TypeError: unhashable type:  'bytearray'` while calling `RESP_DECODING.get()` with data from libusb.

Not having any clue if it was normal or not I tried running scripts from the `misc` folder, which failed for this reason and other ones.

The proposed commits fix the problem with `RESP_DECODING.get()` and some other small fixes in `misc` scripts about Python3 or deprecated functions from inkscape-silhouette.

Great work otherwise, this extension works like a charm for now with my new (although old-ish) device !

Environment details:
- OSX x86_64
- libusb 1.0.24 (from brew)
- Python 3.9.5 (from brew)
  - libusb1==1.9.2 (from pip)
  - pyusb==1.1.1 (from pip)

### DIFF
--- a/misc/arrow_test.py
+++ b/misc/arrow_test.py
@@ -22,10 +22,10 @@ print("status=%s" % dev.wait_for_ready(20))
 
 tmp_fwd=85	# enough to show 20mm of the latest drawing on the far side of the device.
 
-dev.setup(media=113, pressure=0, trackenhancing=True, return_home=False)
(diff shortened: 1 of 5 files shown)


t0b3:
@fcharlier thanks for your contribution. hopefully it works now well on MacOS...

PR #156, opened 2021-09-06
chore: Restore silhouette_multi to operation

### LINKED ISSUE #155: Status of "multi action" variant of plugin?
I was just trying to set up a quick way to install the extension just for my user account in Linux (for testing during development work) and discovered that there already is an undocumented `make install_local` that does this. Excellent. But then I noticed that it also installs a "Silhouette Multi Action" variant of the plugin as well.

This variant does not seem to be documented anywhere, although it is mentioned in one issue, #70. As near as I can guess, it seems maybe that it is intended to allow different parameters for different color strokes in the SVG file. As I often want to score and cut in one pass, I would totally use such a feature.

But even after `make install_local` (from current HEAD), the "Silhouette Multi Actions..." extension doesn't do anything except put up a dialogue with a "dry run" checkbox and "Cancel" and "Apply" buttons. Hitting "Apply" just causes an alarm ding, with no display or activity. So the extension seems broken? Any thoughts on the feasibility and desirability of restoring it?

I will give it a bit of a try now since this would be a very useful feature for me, but if anyone reading this has thoughts or guidance on getting silhouette_multi working again, I'd appreciate hearing.

gwhitney:
It appears that silhouette_multi was never updated for compatibility with
  Inkscape v. 1.0  and/or python3. This commit does so.

  Unfortunately, as of Inkscape v. 1.1, inkex can't be imported when
  sys.stdout has been closed. Therefore, the code in sendto_silhouette must
  take extra care and precaution with where its output is being directed.

  Also (to aid the process of debugging), this commit enhances the logging
  facility in silhouette_multi to produce something even when standard output
  and standard error are closed so that inkscape can continue its operation.
  In conjunction with that change, the commit adds initial configuration
  options to retain control until silhouette_multi completes, and to perform
  more verbose logging.

  This commit includes one minor behavior enhancement: you can now specify
  the behavior for svg items that have no color assigned (formerly they were
  ignored entirely by silhouette multi).

  Finally, the commit extends the README documentation to cover
  silhouette_multi and its installation, to increase awareness of it.

  Resolves #155

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ Install Inkscape
 Install requirements
 * `sudo apt install python3-usb`
 
-Install inscape-silhouette
+Install inkscape-silhouette
 * https://github.com/fablabnbg/inkscape-silhouette/releases
   Scroll down to Downloads and click on the *.deb file.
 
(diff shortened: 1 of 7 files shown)


gwhitney:
Sorry, forgot that f""-string syntax is not (yet) allowed in inkscape-silhouette for the sake of Python2 compatibility. Will revise and push an amended commit.

t0b3:
great contribution - big thx!

PR #158, opened 2021-09-06
feat: allow inkscape_silhouette to append to the log/dump files

gwhitney:
(As a selectable option to the previous behavior, of overwriting
  the dump/log if they already existed. This is primarily
  motivated by #156, because it multi operation it makes most
  sense to append to the log and/or command file. However, it can
  also be a standalone change to the basic inkscape_silhouette, so
  I thought I should submit it separately.)

  In testing this, I found there was no way to select the Cameo
  cutting mat that I had loaded in my machine, so I added a couple
  of new mat types.

  That in turn meant I had to find the commands that needed to be
  issued for these mat types. So I indulged in a bit of USB packet
  capture, which I recorded in Commands.md along with a
  comprehensive list of all of the commands that appear to have
  been used/documented to date.

  I recognize this is in some sense a mashup of what might most
  properly be two commits or even PRs: add the mats (and the 
  command docs) on the one hand, and allow log appending on the
  other. I was hoping that it would be OK since they're both pretty
  minor and I got onto the mats in trying to test my change to the
  logs -- it just caught me by surprise that the larger mats were
  not already in the extension. But if the maintainers would like me
  to split this in two, please let me know.

  Finally, whichever of this and #156 is merged first, the other will
  require a tiny bit of manual merging, as they both affect the
  opening of the log file in silhouette_multi.py. I will be happy to
  do that, as soon as one is merged.

### DIFF
--- a/Commands.md
+++ b/Commands.md
@@ -15,6 +15,178 @@ Resources
  * https://github.com/fablabnbg/inkscape-silhouette/blob/master/silhouette/Graphtec.py#L305-L419
  * https://github.com/Skrupellos/silhouette/blob/master/decode
 
+Command Summary
+---------------
+
+This section summarizes in ASCII order the known and observed commands understood by one or more
+models in the Silhouette Cameo line (along with some Graphtec models not sold under the
+Silhouette brand). The original source for most of these commands is the GPGL reference for the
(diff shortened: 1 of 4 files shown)


t0b3:
You can resolve the conflicts...

gwhitney:
Should be OK now.

PR #167, opened 2022-01-21
Commands updated according to command set manual

Snow4DV:
I've got a graphtec plotter (MP4300) on my hands and there was a full GP-GL command set manual that may help you.
Note that it's quite old (late 80-s - early 90-s) so it is missing lots of stuff probably added to newer machines
but still makes many commands more clear:
https://github.com/Snow4DV/graphtec-gp-gl-manual/raw/main/Manual.pdf (shrinked to 20mb)
https://drive.google.com/file/d/1B3RUKR-3AUZ70lLAzPTq8G_it09QH3mp/view?usp=sharing (init 60mb)


Snow4DV:
//delete

Issue #130, opened 2021-04-12
Install using .deb not working on Ubuntu 20.04 with inkscape 1.02


concretedog:
First of all, many thanks for  your work this amazing extension, its a fabulous tool. 

Following the instructions for install on ubuntu 20.04 I downloaded the .deb (having installed python3 -usb) and installed and it didn't work/appear in inkscape 1.02. I resolved this by following the later install for other debian based distros and manually unzipping and adding the folder and two files to .config/inkscape/extensions, works perfectly now. 

Thanks again.

t0b3:
@concretedog did you get any error messages? or did the deb install correctly? [need more info] 
what would need to be changed? 
please feel free to make a PR...

concretedog:
So I tired to install using the instructions on the readme and double clicked the downloaded .deb, it launched into the Ubuntu Software application and appeared to install correctly, no error messages, then opening inkscape there was no sign of the extension, using the other method of copy pasting from the downloaded zip worked though. I am definitely not skilled enough to make a PR :) Just wanted to feedback my experience so fee free to close.

t0b3:
a) what is the output of `dpkg -L inkscape-silhouette | grep /usr/share/inkscape/extensions/`?
b) which directory is in Inkscape `Menu -> Edit -> Preferences -> System -> Inkscape extensions`?

concretedog:
The output of dpkg is 
`/usr/share/inkscape/extensions/sendto_silhouette.inx
/usr/share/inkscape/extensions/sendto_silhouette.py
/usr/share/inkscape/extensions/silhouette
/usr/share/inkscape/extensions/silhouette/Geometry.py
/usr/share/inkscape/extensions/silhouette/Graphtec.py
/usr/share/inkscape/extensions/silhouette/Strategy.py
/usr/share/inkscape/extensions/silhouette/StrategyMinTraveling.py
/usr/share/inkscape/extensions/silhouette/__init__.py
/usr/share/inkscape/extensions/silhouette/__pycache__
/usr/share/inkscape/extensions/silhouette/__pycache__/Geometry.cpython-38.pyc
/usr/share/inkscape/extensions/silhouette/__pycache__/Graphtec.cpython-38.pyc
/usr/share/inkscape/extensions/silhouette/__pycache__/Strategy.cpython-38.pyc
/usr/share/inkscape/extensions/silhouette/__pycache__/StrategyMinTraveling.cpython-38.pyc
[...]

PR #173, opened 2022-03-05
Install extensions as console_scripts

twmr:
Since a `setup.py` file exists in the src tree, it should also be used for the installation, which is currently not the case if one follows the instructions in the README. The README says that the requires files should be copied into the inkscape `extensions` directory. IMO the files in the `silhouette` directory should be properly installed and executables of the two extensions should be created, which should be used in the `inx` files.

This is done in this PR. Furthermore the pytest related code from setup.py was removed, because
the `test` subcommand of `setup.py` is deprecated.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -75,6 +75,9 @@ jobs:
         python -m pip install --upgrade pip
         pip install -r requirements.txt
         pip install pytest
+    - name: install pkg
+      run: |
+        pip install .
     - name: test
       env:
         PYTHONWARNINGS: default
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,3 @@
 *.pyc
+build/
+/inkscape_silhouette.egg-info/
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ PREFIX?=/usr
 DISTNAME=inkscape-silhouette
 EXCL=--exclude \*.orig --exclude \*.pyc
 ALL=README.md *.png *.sh *.rules *.py *.inx examples misc silhouette
-VERS=$$(python3 ./sendto_silhouette.py --version)
+VERS=$$(python3 ./silhouette/send.py --version)
 
 ## echo python3 ./sendto_silhouette.py
 # 'module' object has no attribute 'core'
@@ -43,15 +43,13 @@ install:
 
 install-local:
 	mkdir -p $(DESTLOCAL)
-	# CAUTION: cp -a does not work under fakeroot. Use cp -r instead.
-	cp -r silhouette $(DESTLOCAL)
-	install -m 755 -t $(DESTLOCAL) *silhouette*.py
+	pip install .
 	install -m 644 -t $(DESTLOCAL) *.inx
 
 tar_dist_classic: clean
 	name=$(DISTNAME)-$(VERS); echo "$$name"; echo; \
 	tar jcvf $$name.tar.bz2 $(EXCL) --transform="s,^,$$name/," $(ALL)
-	grep about_version ./sendto_silhouette.inx 
+	grep about_version ./sendto_silhouette.inx
 	@echo version should be $(VERS)
 
 tar_dist:
--- a/README.md
+++ b/README.md
@@ -135,8 +135,13 @@ sudo make install   # OR: make install-local  # latter installs only for this us
 #### Silhouette inkscape extension itself
 
 * Download https://github.com/fablabnbg/inkscape-silhouette/archive/master.zip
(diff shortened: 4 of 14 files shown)


t0b3:
While the current default installation method is to install the files as a Inkscape Extension...
this PR  is designed to IIUC: 
1. modernize setup.py (drop test, etc.)
2. change install to console_script for standalone use
3. adapt instructions and installation scripts for new installation method

I'd like to concentrate this PR to changes [2] and [3]. 
- Thus I'd prefer if you can open a separate PR for [1] to focus on PEP517, PEP518 and PEP621 `setup.py` incl. `setup.cfg` and `pyproject.toml`. There are some great examples i.e. https://github.com/dib-lab/screed/pull/83/files or https://github.com/RDFLib/sparqlwrapper/pull/196/files. It would be great to have the standalone install capabilities for `pip install` in a current state... and eventually have it documented in some Readme chapter for standalone use w/o inkscape.

__

May I ask: will the changes [part 2] 
- work to provide the extension in one bundle to consume via inkscape extension manager (see `inkscape menu -> extensions -> manage -> install` and https://inkscape.org/develop/extensions/)? 
- or would they involve to  install two separate parts (inkscape extension, python package) *and keep them in sync subsequently*? 
image

It would be great to know how these changes support solving [some nasty snap related path issues for inkscape](https://github.com/fablabnbg/inkscape-silhouette/issues/130) in the future i.e. via deploying the extension for the extension manager... Thanks for clarification.
__

PS: In case we choose to abandon the current dual installation possibilities and switch to solely pip install method then we need to ensure it will apply everywhere in the right context (system vs. user: i.e. `pip install --user`) and for packaging for all platforms (linux, mac, win) and adapt installation instructions as well. This would need to consider the build-deb workflow, the relevant parts of the installation related files and workflows as well as more chapters of Readme.
https://github.com/fablabnbg/inkscape-silhouette/blob/5a19eb4526d96371ff61683aa6e31f5bf6c5c537/.github/workflows/build-deb.yml#L38
https://github.com/fablabnbg/inkscape-silhouette/blob/5a19eb4526d96371ff61683aa6e31f5bf6c5c537/Makefile#L37
https://github.com/fablabnbg/inkscape-silhouette/blob/5a19eb4526d96371ff61683aa6e31f5bf6c5c537/install_osx.py#L18
https://github.com/fablabnbg/inkscape-silhouette/blob/5a19eb4526d96371ff61683aa6e31f5bf6c5c537/distribute/distribute.sh#L24
https://github.com/fablabnbg/inkscape-silhouette/blob/5a19eb4526d96371ff61683aa6e31f5bf6c5c537/README.md?plain=1#L40-L42
https://github.com/fablabnbg/inkscape-silhouette/blob/5a19eb4526d96371ff61683aa6e31f5bf6c5c537/README.md?plain=1#L112
etc.
...

twmr:
@t0b3 thx a lot for your feedback!  You are right that my PR basically consists of 3 parts (setup.py, console-scripts, installation scripts). 

> I'd like to concentrate this PR to changes [2] and [3].

Fine by me. I'll remove the setup.py changes and create a separate PR for them once parts 2 and  3 are merged.

> will the changes [part 2] work to provide the extension in one bundle to consume via inkscape extension manager?

I think it is possible to extend setup.py to install the inx file to the desired location e.g. into the home directory of the user (`~/.config/inkscape/extensions`). But I don't know what is required to create a bundle that can be installed via the inkscape extension manager.

I just had a look at two extensions that are part of the inkscape-extension-store:

https://gitlab.com/inkscape/extras/inkscape-import-clipart
https://github.com/raniaamina/inkporter/tree/inkporter-gui

but none of them use setup.py to install the extension. I'll look further - maybe I can find an extension that has already solved the installation/deployment problem.

twmr:
In order to simplify and generalize the installation instructions, we could bundle all files (inkscape extension + python library) into a python pkg, which can be installed using `pip`.  Part of this package should be a new console script that copies the extension related files to a desired destination location. So, the user would have to install the pkg using pip and then run this inkscape-extension-copy script.

twmr:
@t0b3  I've removed the setup.py changes from this PR as you suggested. In a separate PR I'll create a new console script that allows the users copy the inx files into the desired extension directory. This would allows us to simplify the installation instructions. WDYT?

t0b3:
Which use case will this PR contribute to? Is this PR suggesting to install the inkscape extension via `pip -> script -> inkscape extension`? 

If yes
- how would you like to avoid this becoming a new point of failure to keep the versions and paths of INX and PY files in sync in all those installations out there... 
- how would this affect the update process? is manual intervention needed? ...
- how would this affect selection of installation paths? is user knowledge of paths needed? ...

and for which OS and which steps do you feel do the installation instructions sound complicated?

PR #182, opened 2022-04-23
Fix handling of duplicates

ktosiek:
Duplicated nodes had the document transform applied twice, once in the inherited extra_transform and then again at the top of recursivelyTraverseSvg.


t0b3:
nice catch!

Issue #172, opened 2022-03-05
silhouette_multi invoking sendto_silhouette improperly when "Override cutter model" selected


gwhitney:
If I use silhouette multi and a value of "Override cutter model" as anything besides `-- as detected --`, the cut fails. The difficulty appears to be that then the value of the `--force_hardware` option has a space in it, but this space is not quoted or escaped in any way, leading to `sendto_silhouette` not properly parsing the arguments.

I'd suggest a workaround of changing all of the spaces to underscores in the "value" fields of the force_hardware section of `sendto_silhouette.inx` and then making the identical changes to the "name" fields of the DEVICE array in `sihouette/Graphtec.py`. I would supply a PR myself, but my repository is out of whack with the github one since I submitted the PR #161 (and I keep hoping that will happen at some point).

t0b3:
@gwhitney in the mean time while waiting for #161 you can still sync & update your master branch and thank make a new branch and PR that one...

PR #188, opened 2022-05-23
feat(multi): Show the return code in case command fails

gwhitney:
First, thanks so much for renaming the primary branch to main. I will also get to work on #172.

This is a small code change I have had sitting in my clone that I found very helpful in debugging on occasion; it has no effect except when silhouette_multi (which I now use almost exclusively) fails.

### DIFF
--- a/silhouette_multi.py
+++ b/silhouette_multi.py
@@ -826,14 +826,15 @@ class SilhouetteMulti(inkex.Effect):
 
     def run_commands_with_dialog(self, commands):
         for i, command in enumerate(commands):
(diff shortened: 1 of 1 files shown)


gwhitney:
Oops, ignore this PR, silhouette_multi is again not working at all. (I should have tested again after the merge before submitting this change, stupidly thought it was so small what could go wrong.) Will revisit once silhouette_multi is working again.

t0b3:
never mind. feel free to contribute to get multi back up again. your work on that is welcome... :smile:

PR #193, opened 2022-05-29
fix(multi): Remove whitespace from hardware device names

### LINKED ISSUE #172: silhouette_multi invoking sendto_silhouette improperly when "Override cutter model" selected

gwhitney:
When silhouette_multi calls sendto_silhouette, if the force_hardware parameter
  has been set, it places the name of the device being forced on the command
  line. If that name contains any whitespace, the command-line parsing of
  sendto_silhouette fails. This commit resolves the situation by replacing
  all whitespace in device names with underscores.

  Resolves #172.

(I promise the next thing I will work on after this is at least some sort of test for silhouette_multi.)

### DIFF
--- a/sendto_silhouette.inx
+++ b/sendto_silhouette.inx
@@ -134,18 +134,18 @@ Minimal Traveling (no reverse): Like fully optimized but respect original orient
       <!-- CAUTION: keep hardware list in sync with silhouette/Graphtec.py -->
       <param name="force_hardware" type="enum" _gui-text="Override cutter model to:">
 	<item value="DETECT">-- as detected --</item>
-	<item value="Silhouette Portrait">Silhouette Portrait</item>
-	<item value="Silhouette Portrait2">Silhouette Portrait 2</item>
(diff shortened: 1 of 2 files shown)


t0b3:
@gwhitney thank you for all your contributions. if you find some sort of tests, this would be useful. the greatest thing would be some sort of ui emulation tests, however I don't know if that's feasible...

PR #197, opened 2022-06-21
Update silhouette-udev.rules

mfraser:
Remove invalid OPTIONS keys 'lastrule' and 'last_rule'


mfraser:
This no longer produces an error, but I am unable to test.

Issue #157, opened 2021-09-06
Provide a test or tests for silhouette_multi


gwhitney:
Presumably one of the reasons that silhouette_multi fell into an inoperable status is that there are no tests for it in `test/test_run.py`.

Is it considered a core part of inkscape-silhouette? Would the maintainers like some sort of a test? Having just done a fair amount of work to get it running again, I'd be inclined to write a text except given that it uses wxPython internally to put up its own gui, I'm at a bit of a loss how to test it in a framework like test_run.py.

Any thoughts/pointers, let me know, and I will try to work up something. Should I add a no_gui command line argument and use a pre-pickled setting just to test the module loads and does the color separation properly?

t0b3:
this would be very welcome, so we can ensure future contributions are tested against the some basic cases of silhouette_multi. At least some `no_gui` cases would be good. I neither have experience with wxPython tests.

Maybe this helps as a starting point:
- https://wiki.wxpython.org/Unit%20Testing%20with%20wxPython
- https://wxpython.org/Phoenix/docs/html/wx.UIActionSimulator.html

t0b3:
@gwhitney this would be very helpful to avoid unintentional breaking of silhouette_multi action...

gwhitney:
OK, I will start with a no_gui flag so that at least its operation can be tested.

gwhitney:
All right, I have finally managed in the PR just above to add options to specify the pickle file with the settings and an option to disable the GUI. This allows the first test of silhouette_multi (included in the PR) that should ensure that it doesn't completely get messed up by other future code changes. (I still need to iron out the tests on some of the systems; it's working, but there are spurious differences in stderr output.)

It is a somewhat limited test, however, in that a noticeable amount goes on in loading the gui, not least of which is the parsing of the .inx files. In fact, the last time it broke and I repaired it, it was that parsing which broke, and that is not being tested at all by the no GUI test.

To go further, I wasn't sure what the testing environment was like -- is wx even available on the test bench? Since the tests are run headless, what will happen if wx frames are constructed? If wx is available and it is safe to make frames, then I can add a test which runs with the GUI but then rather that running the event loop, just immediately exits with the "run callback". That would definitely cover more of the code. Let me know if that seems feasible with the way the testing is performed, and if so I would be willing to submit another PR with such a test. Thanks for letting me know.

gwhitney:
[...]

PR #203, opened 2022-07-11
Test multi nogui

gwhitney:
This PR adds a "-p" option to silhouette_multi.py to specify the pickle file of stored settings, refactors all of the color separation settings processing to separate it from the wx gui code and moves the wx code into a separate file, and adds a "-g" option to control whether the GUI is used. (If not, it simply loads the pickle file and generates the commands to run based on the default
settings therein.)

It then uses these new options to add a test that silhouette_multi.py runs without the GUI and generates an appropriate color separation.

This at least partially addresses #157.


gwhitney:
Aha the failed tests are due to things like newline versus carriage-return and newline when comparing the actual and proper output, and some extra stuff written to stderr on some of the systems. I will fix and push another commit.

t0b3:
@gwhitney thanks for your great work. unfortunately I didn't manage to get an in depth look at it yet.
may I ask you to adapt some minor things?
- add the new multi related tests in a separate `run_multi.py` file for better separation and easier overview
- move the data related to the new tests in a subfolder within `test` to keep the related data together

this would be helpful

gwhitney:
Done as requested.

t0b3:
LGTM

PR #220, opened 2022-12-24
Prevent running if preview requested & matplotlib not installed

CarlosGS:
The machine started running directly, even when preview was requested, because matplotlib was not installed yet.
Simple fix, now the preview function returns True (cancel job) in that case.


t0b3:
@CarlosGS Thank you for bringing up this issue :+1: I took the opportunity to handle different return values.
Feel free to reopen or bring up a new PR in case there are issues left.

Issue #224, opened 2023-01-13
Installation on Freebsd 13.1 fails - option -t not supported


pbpjackd:
`sudo make install`
 
results in:

```
mkdir -p /usr/share/inkscape/extensions
cp -r silhouette /usr/share/inkscape/extensions
install -m 755 -t /usr/share/inkscape/extensions *silhouette*.py
install: illegal option -- t
usage: install [-bCcpSsUv] [-f flags] [-g group] [-m mode] [-o owner]
               [-M log] [-D dest] [-h hash] [-T tags]
               [-B suffix] [-l linkflags] [-N dbdir]
               file1 file2
       install [-bCcpSsUv] [-f flags] [-g group] [-m mode] [-o owner]
               [-M log] [-D dest] [-h hash] [-T tags]
               [-B suffix] [-l linkflags] [-N dbdir]
               file1 ... fileN directory
       install -dU [-vU] [-g group] [-m mode] [-N dbdir] [-o owner]
               [-M log] [-D dest] [-h hash] [-T tags]
               directory ...
*** Error code 64

Stop.
make: stopped in /tmp/inkscape-silhouette-main
```

t0b3:
@pbpjackd thank you for reporting this probably long standing issue on FreeBSD. can you test and report back whether a Makefile with such instructions would work on FreeBSD? any porting contributions are welcome :+1: 

```makefile
install:
	mkdir -p $(DEST)
	# CAUTION: cp -a does not work under fakeroot. Use cp -r instead.
	cp -r silhouette $(DEST)
	install -m 755 *silhouette*.py $(DEST)
	install -m 644 *.inx $(DEST)
	mkdir -p $(UDEV)/rules.d
	install -m 644 silhouette-udev.rules $(UDEV)/rules.d/40-silhouette-udev.rules
	install -m 644 silhouette-icon.png $(DEST)
	install -m 644 silhouette-udev-notify.sh $(DEST)
```

t0b3:
see 439a51acd488905f2e61ee4261861aae48ad43f4

PR #225, opened 2023-01-18
Installation on Freebsd 13.1 fails - option -t not supported  #224

pbpjackd:
fix FreeBSD installation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,7 @@ DISTNAME=inkscape-silhouette
 EXCL=--exclude \*.orig --exclude \*.pyc
 ALL=README.md *.png *.sh *.rules *.py *.inx examples misc silhouette
 VERS=$$(python3 ./sendto_silhouette.py --version)
+OSNAME != uname
 
 ## echo python3 ./sendto_silhouette.py
 # 'module' object has no attribute 'core'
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -34,19 +35,35 @@ install:
 	mkdir -p $(DEST)
 	# CAUTION: cp -a does not work under fakeroot. Use cp -r instead.
 	cp -r silhouette $(DEST)
+
+ifeq ($(OSNAME), FreeBSD)
```
t0b3, line 39: let's drop `-t` for all platforms ...

Makefile
```diff
@@ -34,19 +35,35 @@ install:
...
+endif
 
 install-local:
 	mkdir -p $(DESTLOCAL)
 	# CAUTION: cp -a does not work under fakeroot. Use cp -r instead.
 	cp -r silhouette $(DESTLOCAL)
-	install -m 755 -t $(DESTLOCAL) *silhouette*.py
+
+ifeq ($(OSNAME), FreeBSD)
+	install -m 755 *silhouette*.py $(DESTLOCAL)
```
t0b3, line 61: let's drop `-t` for all platforms ...

README.md
```diff
@@ -110,7 +110,7 @@ cd /tmp
 wget -c "https://github.com/fablabnbg/inkscape-silhouette/archive/main.zip"
 unzip main.zip
 cd inkscape-silhouette-main
-sudo make install   # OR: make install-local  # latter installs only for this user
```
t0b3, line 113: hmm, I don't think `gmake` is not required though. with the above it Makefile becomes compatible with the FreeBSD variant of `make` , won't it?

distribute/distribute.sh
```diff
@@ -51,6 +51,18 @@ then
...
 
+echo "****************************************************************"
+echo "FreeBSD Version: For Building you must have gmake installed"
+echo "Build FreeBSD Version (Y/n)?"
+read answer
+if [ "$answer" != "n" ]
+then
+  mkdir -p bsd/files
+  cp -a $name/* bsd/files
+  (cd bsd && sh ./dist.sh $name $VERSION)
```
t0b3, line 62: bsd/dist.sh would be missing

distribute/distribute.sh
```diff
@@ -51,6 +51,18 @@ then
 fi
 
 
+echo "****************************************************************"
```
t0b3, line 54: this whole script is meant for packaging some `deb` and the like. 
how do you package this for FreeBSD? and how would it's content differ from a Linux package?
I'd prefer to handle this in another PR though...

t0b3 (changes requested):
@pbpjackd thank you for your contribution. Can you address the points found in the code review and test again yourself? thank you :+1:

README.md
pbpjackd: Yes

distribute/distribute.sh
pbpjackd: Packaging is not needed at this point I think. The reason I wanted to use your logic their was because I need to do this somewhere:

`sed -i -e 's@"/lib/udev/@"/usr/local/share/inkscape/extensions@' files/silhouette-udev.rules`

and also override UDEV and PREFIX in the makefile to match FreeBSD.

`make install UDEV=/usr/local/lib/udev PREFIX=/usr/local`

Any suggestion where to put this ?

t0b3: does `notify-send` indeed work and can it be triggered through `udev` on FreeBSD?
FYI: I don't use the whole udev related stuff neither on my system.

t0b3:
see 439a51acd488905f2e61ee4261861aae48ad43f4 
@pbpjackd feel free to come up with another PR in case there is still something missing :+1:

PR #233, opened 2023-04-05
Improve out-of-the-box support of Silhouette Cameo 2

neomilium:
This PR contains minor improvements to make Silhouette Cameo 2 usable out-of-the-box under Debian-based systems.

First, it add a new `udev` rule to setup Cameo 2 model.

Next, it removes the hardcoded margins for Cameo 2 model.
These margins were hardcoded to 9mm on left, 1mm on top… which require to set a X-offset to -9 and Y-offset to -1 at cutting time.
Please note I remove margins and made some tests with a plain success but I did not really understand if it has side effect.

Finally, I turn the example with registration marks in a distributed template.


t0b3:
@neomilium well done :tada: 
missed to ask: can you add LOCAL_TEMPLATES dir as well for local-install ?

neomilium:
> @neomilium missed to ask: can you add LOCAL_TEMPLATES dir as well for local-install ?

You can find expected code in #234

PR #249, opened 2023-08-13
sento_silhouette: logfile param defaults to /tmp/inkscape-silhouette.log

mofosyne:
Still trying to understand how inx works, but I think this would be how to add /tmp/inkscape-silhouette.log as the default logfile path.

This consistency if correct would make bug filing easier for newbies.


t0b3:
INX value works like this. :+1: 
Can you test if this default value works for Win and Mac?

mofosyne:
Don't have windows or mac, but did some digging and this seems like the best approach.

Added a function called get_default_logfile_path() to auto create a logfile path if non is given.

----

image

```
~$ cat /tmp/inkscape-silhouette.log 
Inkscape 1.3 (1:1.3+202307231459+0e150ed6c4)
Inkex: 1.3.0
Inkscape-Silhouette: 1.28
Path: /usr/share/inkscape/extensions/sendto_silhouette.py
Python: /usr/bin/python3
Version: 3.10.12 (main, Jun 11 2023, 05:26:28) [GCC 11.4.0]
Platform: linux
Arguments: sendto_silhouette.py --active-tab=logdump --x_off=0.00 --y_off=0.00 --cuttingmat=cameo_12x12 --toolholder=1 --tool=default --media=300 --speed=0 --pressure=0 --depth=-1 --preview=true --dashes=false --autocrop=false --bbox-only=false --multipass=1 --reversetoggle=false --endposition=start --end_offset=0.0 --regmark=false --regsearch=true --regwidth=190.0 --reglength=277.0 --regoriginx=10.0 --regoriginy=10.0 --wait_done=false --sharpencorners=false --sharpencorners_start=0.1 --sharpencorners_end=0.1 --overcut=0.5 --strategy=zorder --orient_paths=natural --fuse_paths=true --sw_clipping=true --create_log=true --log_paths=false --append_logs=false --inc_queries=false --dry_run=false --force_hardware=DETECT --bladediameter=0.9 /tmp/ink_ext_XXXXXX.svgG3SA91
Preview aborted.
 done. 0 min 0 sec
```

mofosyne:
This is ready, just needs merging

sendto_silhouette.py
```diff
@@ -8,7 +8,7 @@
 __version__ = "1.28"     # Keep in sync with sendto_silhouette.inx ca line 179
 __author__ = "Juergen Weigert <juergen@fabmail.org> and contributors"
 
-import sys, os, time, math, operator
+import sys, os, time, math, operator, platform
```
t0b3: `platform` isn't needed anymore

sendto_silhouette.py
```diff
@@ -68,6 +68,8 @@
     SvgDocumentElement.viewport_height = property(lambda self: convert_unit(self.get("height"), "px") or self.get_viewbox()[3])
     SvgDocumentElement._base_scale = lambda self, unit="px": (convert_unit(1, unit) or 1.0) if not all(self.get_viewbox()[2:]) else max([convert_unit(self.viewport_width, unit) / self.get_viewbox()[2], convert_unit(self.viewport_height, unit) / self.get_viewbox()[3]]) or convert_unit(1, unit) or 1.0
 
+# Default Logfile Filename
+LOGFILE_DEFAULT_NAME = "inkscape-silhouette.log"
```
t0b3: LOGFILE_DEFAULT_NAME = "silhouette.log"

using this as a default value for better compatibility with existing code

sendto_silhouette.py
```diff
@@ -135,6 +138,9 @@ def add_arguments(self, pars):
         pars.add_argument("-D", "--depth",
                 dest = "depth", type = int, default = -1,
                 help="[0..10], or -1 for media default")
+        pars.add_argument("--disable_logging",
+                dest = "disable_logging", type = Boolean, default = False,
+                help="Disable logging to file")
```
t0b3: I'd rather omit yet another parameter but to log always (which won't hurt AFAICT)

t0b3 (commented):
this looks elegant and seems to work on any platforms.

can you ensure it works on different platforms by introducing some kind of test? 
i.e. check for `self.default_logfile_path` and assert you can write a file to this path by appending another test?

PR #252, opened 2023-08-14
skip registration mark and print layer

mofosyne:
This helps improve the UX experience for users, as they would not need to remember to hide the print or registration mark layer

### DIFF
--- a/sendto_silhouette.py
+++ b/sendto_silhouette.py
@@ -306,6 +306,17 @@ class SendtoSilhouette(EffectExtension):
         for node in aNodeList:
             # Ignore invisible nodes
             if isinstance(node, BaseElement):
+                # Check if layer name is referring to registration mark or print layer
+                LAYER_LABEL_ATTR_KEY = '{http://www.inkscape.org/namespaces/inkscape}label'
+                if LAYER_LABEL_ATTR_KEY in node.attrib:
+                    layer_label_name = node.attrib[LAYER_LABEL_ATTR_KEY]
+                    if "regmark" in layer_label_name.lower():
+                        self.report(f"layer '{str(layer_label_name)}' is a registration mark layer - skipped", 'log')
+                        continue
+                    if "print" in layer_label_name.lower():
+                        self.report(f"layer '{str(layer_label_name)}' is a print layer - skipped", 'log')
+                        continue
+
                 # try:
                 #     # Inkex 1.2: `cascaded_style()` considers CSS (has bad performance!!)
                 #     get = node.cascaded_style().get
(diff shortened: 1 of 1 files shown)


mofosyne:
Have not updated readme for it, but may consider if needed.

sendto_silhouette.py
```diff
@@ -306,6 +306,17 @@ def recursivelyTraverseSvg(self, aNodeList,
         for node in aNodeList:
             # Ignore invisible nodes
             if isinstance(node, BaseElement):
+                # Check if layer name is referring to registration mark or print layer
```
t0b3: the whole check would only be needed on `Group`s, right? if so, move to Group logic below... this will help to speed things up

sendto_silhouette.py
```diff
@@ -306,6 +306,17 @@ def recursivelyTraverseSvg(self, aNodeList,
...
+                LAYER_LABEL_ATTR_KEY = '{http://www.inkscape.org/namespaces/inkscape}label'
+                if LAYER_LABEL_ATTR_KEY in node.attrib:
+                    layer_label_name = node.attrib[LAYER_LABEL_ATTR_KEY]
```
t0b3: you may access the layer's label through `node.label` property 
see https://gitlab.com/inkscape/extensions/-/blob/master/inkex/elements/_base.py#L535-538

thus simply use
```
if node.layer:

**Review by t0b3 — CHANGES_REQUESTED** [review]

**Comment by mofosyne** [comment]
Okay, correction done

File: sendto_silhouette.py [hunk 3]
```diff
@@ -340,6 +329,15 @@ def recursivelyTraverseSvg(self, aNodeList,
                     my_transform = parent_transform @ node.transform
 
             if isinstance(node, Group):
+                # Check if layer name is referring to registration mark or print layer
+                LAYER_LABEL_ATTR_KEY = '{http://www.inkscape.org/namespaces/inkscape}label'
```
t0b3: `LAYER_LABEL_ATTR_KEY` isn't needed anymore, correct?

sendto_silhouette.py
```diff
@@ -340,6 +329,15 @@ def recursivelyTraverseSvg(self, aNodeList,
...
+                if node.label:
+                    if "regmark" in node.label.lower():
+                        self.report(f"layer '{str(node.label)}' is a registration mark layer - skipped", 'log')
```
t0b3: `node.label` already returns string type, no need to call `str()`

sendto_silhouette.py
```diff
@@ -340,6 +329,15 @@ def recursivelyTraverseSvg(self, aNodeList,
...
+                        continue
+                    if "print" in node.label.lower():
+                        self.report(f"layer '{str(node.label)}' is a print layer - skipped", 'log')
```
t0b3: `node.label` already returns string type, no need to call `str()`

t0b3 (changes requested):

mofosyne:
Good catch. All resolved.

PR #251, opened 2023-08-14
render_silhouette_regmarks: render rego mark (From miLORD1337)

mofosyne:
Found [miLORD1337](https://github.com/miLORD1337) post in https://github.com/fablabnbg/inkscape-silhouette/issues/79 where he was proposing his work.

His code is GPLed and looks suitable for integration into this extension for convenience. For user experience, I've made the width and length parameter optional so that the user only needs to care about the spacing from edge of the document (we can generally assume that the document would be the print size (e.g. A4)... if that's not the case then the user can still override that value anyway.

Next thing we could do is update the reg mark page in the main script to also make these same parameter optional as well.

https://github.com/miLORD1337/silhouette-regmarks

image

PR #253, opened 2023-08-19
sentto_silhouette: Auto calc mark to mark distance. Also set origin offset to 10x10mm spacing default

mofosyne:
Purpose of this is to reduce the mental load of having to think about the document size. Instead they just need to set the regmark origin. The width and length spacing would instead be autocalculated from the document size.

This approach would improve the UX of new users when combined with this PR https://github.com/fablabnbg/inkscape-silhouette/pull/251 to render rego marks, by reducing the number of variables they have to consider and config.

This approach is based on the assumption that the user will always be maximally using the material stock area.
Thus this may not be the golden path for those placing multiple "reg marks" into one card stock.
But in practice this is not usually common, most people would just be printing onto A4 sized paper or sticker (gut feeling).


sendto_silhouette.py
```diff
@@ -555,6 +555,11 @@ def multipassOvercut(self, paths, multipass, reversetoggle, overcut):
 
 
     def effect(self):
+        reg_origin_X = self.options.regoriginx
+        reg_origin_Y = self.options.regoriginy
+        reg_width = self.options.regwidth if self.options.regwidth else int(self.svg.get("width").rstrip("mm")) - reg_origin_X*2
+        reg_length = self.options.reglength if self.options.reglength else int(self.svg.get("height").rstrip("mm")) - reg_origin_Y*2
+
```
mofosyne: It's from https://github.com/fablabnbg/inkscape-silhouette/pull/251 

Looking at it closely... I feel like there's gotta be a more official way to do `self.svg.get("height").rstrip("mm")`

t0b3: It's a perfect plan to detect regmarks from svg WRT #254. 
NB: those regmark IDs are used allover the places and in other templates as well 
see https://github.com/search?q=repo%3Afablabnbg%2Finkscape-silhouette%20id%3D%22regmark-tl%22&type=code 
i.e.
https://github.com/fablabnbg/inkscape-silhouette/blob/af3c9019e948aa56c9a357263eec45a35df18f20/templates/silhouette-cameo-registration-marks-a4.svg#L149-L160

... and of course this early draft is based on several assumptions
- document settings in mm
- equal borders on both sides

which are not necessarily true. thus further work will be needed... :+1: 
in case code repeats for x and y: consider to define a fuction and call it twice :smile:

mofosyne: Updated the render regmark PR to keep to the consistent svg id

```python
REGMARK_LAYERNAME = 'Regmarks'
REGMARK_LAYER_ID = 'regmark'
REGMARK_TOP_LEFT_ID = 'regmark-tl'
REGMARK_TOP_RIGHT_ID = 'regmark-tr'
REGMARK_BOTTOM_LEFT_ID = 'regmark-bl'
REGMARK_SAFE_AREA_ID = 'regmark-safe-area'
REGMARK_NOTES_ID = 'regmark-notes'
```
This will help with future effort to auto read these metadata.

Had a quick look if I want to tackle the retrieval of the mark to mark distance and mark to origin distance from the svg elements. Not sure I want to tackle that myself, I've updated the issue ticket with the observations so future contributors can jump in faster if they want to https://github.com/fablabnbg/inkscape-silhouette/issues/254 .

I would like to just merge this instead as what you are suggesting appears to be outside the scope of this PR.

mofosyne: Opted to use `self.svg.to_dimensional(self.svg.get("width"), "mm")`

t0b3:
you may try something like this :+1: 
```
    length = self.svg.unit_to_viewport(
                        (self.svg.getElementById('regmark-bl').bounding_box(transform=True).y.maximum
                        - self.svg.getElementById('regmark-tr').bounding_box(transform=True).y.minimum),
                    "mm")

**Comment by mofosyne** [comment]
Thanks for the suggestion. Ported the monkeypatch from #251 to here. Left some extra ported function in even when not used yet as I can imagine there may be future case where functions like unit_to_viewport() may be needed

**Review by mofosyne — COMMENTED** [review]
Looks like no issue with integration checks. Monekypatch ported.  Also added shortcircuit.
This is ready to merge which should complete the render regmark integration.

(By the way, don't forget the other PR i made on default logging and Readme, which should help in dev quality of life)

File: sendto_silhouette.py [hunk 2]
```diff
@@ -555,6 +563,11 @@ def multipassOvercut(self, paths, multipass, reversetoggle, overcut):
 
 
     def effect(self):
+        reg_origin_X = self.options.regoriginx
+        reg_origin_Y = self.options.regoriginy
+        reg_width = self.options.regwidth or self.svg.to_dimensional(self.svg.viewport_width, "mm") - reg_origin_X * 2
+        reg_length = self.options.reglength or self.svg.to_dimensional(self.svg.viewport_height, "mm") - reg_origin_Y * 2
```
t0b3: so to auto detect the regmarks and fix #254 this should do... can you test and give your feedback?
```
        try:
            doc_reg_x = self.svg.unit_to_viewport(self.svg.getElementById('regmark-tl').bounding_box(transform=True).x.minimum, "mm")
            doc_reg_y = self.svg.unit_to_viewport(self.svg.getElementById('regmark-tl').bounding_box(transform=True).y.minimum, "mm")
            doc_reg_width = self.svg.unit_to_viewport(self.svg.getElementById('regmark-tr').bounding_box(transform=True).x.maximum), "mm") - doc_reg_x
            doc_reg_length = self.svg.unit_to_viewport(self.svg.getElementById('regmark-bl').bounding_box(transform=True).y.maximum, "mm") - doc_reg_y
        except Exception:
            doc_reg_x = doc_reg_y = doc_reg_width = doc_reg_length = 0
        reg_origin_X = self.options.regoriginx or doc_reg_x
        reg_origin_Y = self.options.regoriginy or doc_reg_y
        reg_width = self.options.regwidth or doc_reg_width or self.svg.to_dimensional(self.svg.viewport_width, "mm") - reg_origin_X * 2
        reg_length = self.options.reglength or doc_reg_length or self.svg.to_dimensional(self.svg.viewport_height, "mm") - reg_origin_Y * 2
```
... maybe there would be some minor additions useful i.e. add a label in inx to explain 0 means autodetection. and possibly go for 0 as default parameters in py and eventually even in inx parameters... and I hope you can do some tests with your hardware?

mofosyne: I like it. Especially since you used `bounding_box().x.minimum` that looks much more elegant than my initial `float(regmark_tr_d.split(" ")[4])` attempt.

I've mostly adopted your approach, but added some extra checks to only autocalc Regmark layer if all the regmarks objects exist as expected. Also cleaned up the constants svg IDs. And added some logging.

I also agree in removing the minimum default checks and added extra notes to INX as well.
Hopefully that's all needed for now.

One other changes that we may want to do, is to print the regmark setting in the preview window so people can double check that the values is the same as their document visually. But that is a separate PR in my opinion if required

mofosyne:
```
cat /tmp/silhouette.log
Inkscape 1.3 (1:1.3+202307231459+0e150ed6c4)
Inkex: 1.3.0
Inkscape-Silhouette: 1.28
Path: /usr/share/inkscape/extensions/sendto_silhouette.py
Python: /usr/bin/python3
Version: 3.10.12 (main, Jun 11 2023, 05:26:28) [GCC 11.4.0]
Platform: linux
Arguments: sendto_silhouette.py --id=path1 --active-tab=logdump --x_off=0.00 --y_off=0.00 --cuttingmat=cameo_12x12 --toolholder=1 --tool=default --media=300 --speed=0 --pressure=0 --depth=-1 --preview=true --dashes=false --autocrop=false --bbox-only=false --multipass=1 --reversetoggle=false --endposition=start --end_offset=0.0 --regmark=true --regsearch=true --regoriginx=0.0 --regoriginy=0.0 --regwidth=0.0 --reglength=0.0 --wait_done=false --sharpencorners=false --sharpencorners_start=0.1 --sharpencorners_end=0.1 --overcut=0.5 --strategy=zorder --orient_paths=natural --fuse_paths=true --sw_clipping=true --logfile=/tmp/silhouette.log --log_paths=false --append_logs=false --inc_queries=false --dry_run=false --force_hardware=DETECT --bladediameter=0.9 /tmp/ink_ext_XXXXXX.svgUC7091
Detected Existing Registration Mark:: mark distance from document: Left=10.0mm, Top=10.0mm; mark to mark distance: X=190.0mm, Y=277.0mm;
Using Registration Mark:: mark distance from document: Left=10.0mm, Top=10.0mm; mark to mark distance: X=190.0mm, Y=277.0mm;
Preview aborted.
 done. 0 min 1 sec
```

mofosyne:
On second thought. Let's just add the rego settings printout to the pyplot anyway

image

sendto_silhouette.py
```diff
@@ -567,6 +584,26 @@ def effect(self):
...
+        try:
+            if self.svg.getElementById(REGMARK_TOP_LEFT_ID) is not None and self.svg.getElementById(REGMARK_TOP_RIGHT_ID) is not None and self.svg.getElementById(REGMARK_BOTTOM_LEFT_ID) is not None:
+                doc_reg_x = self.svg.unit_to_viewport(self.svg.getElementById(REGMARK_TOP_LEFT_ID).bounding_box(transform=True).x.minimum, "mm")
+                doc_reg_y = self.svg.unit_to_viewport(self.svg.getElementById(REGMARK_TOP_LEFT_ID).bounding_box(transform=True).y.minimum, "mm")
+                doc_reg_width = self.svg.unit_to_viewport(self.svg.getElementById(REGMARK_TOP_RIGHT_ID).bounding_box(transform=True).x.maximum, "mm") - doc_reg_x
+                doc_reg_length = self.svg.unit_to_viewport(self.svg.getElementById(REGMARK_BOTTOM_LEFT_ID).bounding_box(transform=True).y.maximum, "mm") - doc_reg_y
+                self.report(f"Detected Existing Registration Mark:: mark distance from document: Left={doc_reg_x}mm, Top={doc_reg_y}mm; mark to mark distance: X={doc_reg_width}mm, Y={doc_reg_length}mm;", 'log')
+        except Exception:
+            doc_reg_x = doc_reg_y = doc_reg_width = doc_reg_length = 0
```
t0b3: doc_reg_x = doc_reg_y = doc_reg_width = doc_reg_length = 0
        if self.svg.getElementById(REGMARK_TOP_LEFT_ID) is not None and self.svg.getElementById(REGMARK_TOP_RIGHT_ID) is not None and self.svg.getElementById(REGMARK_BOTTOM_LEFT_ID) is not None:
            doc_reg_x = self.svg.unit_to_viewport(self.svg.getElementById(REGMARK_TOP_LEFT_ID).bounding_box(transform=True).x.minimum, "mm")
            doc_reg_y = self.svg.unit_to_viewport(self.svg.getElementById(REGMARK_TOP_LEFT_ID).bounding_box(transform=True).y.minimum, "mm")
            doc_reg_width = self.svg.unit_to_viewport(self.svg.getElementById(REGMARK_TOP_RIGHT_ID).bounding_box(transform=True).x.maximum, "mm") - doc_reg_x
            doc_reg_length = self.svg.unit_to_viewport(self.svg.getElementById(REGMARK_BOTTOM_LEFT_ID).bounding_box(transform=True).y.maximum, "mm") - doc_reg_y
            self.report(f"Detected Existing Registration Mark:: mark distance from document: Left={doc_reg_x}mm, Top={doc_reg_y}mm; mark to mark distance: X={doc_reg_width}mm, Y={doc_reg_length}mm;", 'log')

your solution with `if` renders the try ... except obsolete :+1:

sendto_silhouette.inx
```diff
@@ -85,11 +85,14 @@
...
-      <param name="regoriginx" type="float" min="10.0" max="10000" gui-text="Position of regmark from document left [mm]">15</param>
-      <param name="regoriginy" type="float" min="10.0" max="10000" gui-text="Position of regmark from document top [mm]">20</param>
+      <param name="regoriginx" type="float" min="0.0" max="10000" gui-text="Position of regmark from document left [mm]">10</param>
+      <param name="regoriginy" type="float" min="0.0" max="10000" gui-text="Position of regmark from document top [mm]">10</param>
       <label indent="2">Distance of the registration mark edges</label>
+      <label indent="2">Note: If mark offset from document top and left spacing is empty, then it is calculated from Regmark layer.</label>
+      <param name="regwidth" type="float" min="0.0" max="10000" gui-text="X mark to mark distance [mm]">0</param>
+      <param name="reglength" type="float" min="0.0" max="10000" gui-text="Y mark to mark distance [mm]">0</param>
+      <label indent="2">Distance between registration mark edges</label>
+      <label indent="2">Note: If width and length spacing is empty, then it is calculated from Regmark layer or current document size.</label>
```
t0b3: should we clarify?   `0` instead of `empty`

mofosyne:
image

sendto_silhouette.inx
mofosyne: fixed

sendto_silhouette.py
mofosyne: All good. Fixed

mofosyne (commented):
Should be ready now

sendto_silhouette.py
```diff
@@ -567,6 +584,24 @@ def effect(self):
...
+            self.report(f"Detected Existing Registration Mark:: mark distance from document: Left={doc_reg_x}mm, Top={doc_reg_y}mm; mark to mark distance: X={doc_reg_width}mm, Y={doc_reg_length}mm;", 'log')
+
+        # Registration Mark Selection/Calcuation
+        # Note that any user defined values will overrride any auto calcuated values
+        reg_origin_X = self.options.regoriginx or doc_reg_x
+        reg_origin_Y = self.options.regoriginy or doc_reg_y
+        reg_width = self.options.regwidth or doc_reg_width or self.svg.to_dimensional(self.svg.viewport_width, "mm") - reg_origin_X * 2
+        reg_length = self.options.reglength or doc_reg_length or self.svg.to_dimensional(self.svg.viewport_height, "mm") - reg_origin_Y * 2
+        self.report(f"Using Registration Mark:: mark distance from document: Left={reg_origin_X}mm, Top={reg_origin_Y}mm; mark to mark distance: X={reg_width}mm, Y={reg_length}mm;", 'log')
+
```
t0b3: wow great job, just some last mini tweaks before merging...
- [x] just realized we could replace 
  - [x] `x.minimum` with `left`
  - [x] y.minimum with top
  - [x] x.maximum with right
  - [x] y.maximum with bottom
- [x] shift this whole regmark block to it's own function `def detect_regmark()` i.e. next to def initDocScale() and only call it here... this will help to keep the code more structured
  - [x] this involves to rename i.e. `doc_reg_x` to `self.doc_reg_x`, etc. all over the places
:smile:

mofosyne: All good. Fixed

mofosyne (commented):
Alright made the refactoring

sendto_silhouette.py
```diff
@@ -567,6 +606,18 @@ def effect(self):
...
+        # Generated Regmarks from inkscape silhouette embeds ids so we can automatically derive offset calculations from it
+        self.detect_regmark()
+
+        # Registration Mark Selection/Calcuation
+        # Note that any user defined values will overrride any auto calcuated values
+        reg_origin_X = self.options.regoriginx or self.doc_reg_x
+        reg_origin_Y = self.options.regoriginy or self.doc_reg_y
+        reg_width = self.options.regwidth or self.doc_reg_width or self.svg.to_dimensional(self.svg.viewport_width, "mm") - reg_origin_X * 2
+        reg_length = self.options.reglength or self.doc_reg_length or self.svg.to_dimensional(self.svg.viewport_height, "mm") - reg_origin_Y * 2
+        self.report(f"Using Registration Mark:: mark distance from document: Left={reg_origin_X}mm, Top={reg_origin_Y}mm; mark to mark distance: X={reg_width}mm, Y={reg_length}mm;", 'log')
```
t0b3: the very last thing that bothers me: can we pack this part into the function as well?
i.e. include the logic of this 5 lines also withing detect_regmark()

mofosyne: I get the other one, but this seems like diminishing return kind of refactoring/cleanup. But anyway I've added it in. Just waiting for the checks to pass and let's see how it goes!

mofosyne (commented):
All passed!

t0b3:
awesome :smile: :smile:  :smile: 
the last change I ask for is to add a test case, so in case any future changes would break your awesome work, this would be noticed before merging :+1: 

therefore you may 
- [x] use [test/data/detect_regmarks.svg](https://github.com/fablabnbg/inkscape-silhouette/assets/5751047/0bc54b0f-eb2b-472e-bdf9-f6e0e15df9d4) simple svg with scale 2 and regmarks
- [x] add a new testcase (new class) to the bottom of test/test_sendto_silhouette.py
```
class SyncDetectRegmarks(SendtoSilhouetteTest):
    source_file = "detect_regmarks.svg"

    def test_detect_regmarks(self):
        self.e.parse_arguments([self.data_file(self.source_file), "--<your_param_name>=<your_param_value>", "<repeat for more params...>"])
        self.e.load_raw()
        self.e.<your_new_fun_here>()    # i.e. `detect__doc_regmark`

        self.assertEqual(   ## or any other assert* function you'd like...
            self.e.<your_new_var_here>,    # i.e. `doc_reg_width`
            <your_expected_value_here>     # i.e. `190.0`
        )

        [repeat for more var tests if appropriate]

    def test_sync_regmark_settings(self):
        [repeat above pattern...]

```
with this very last step your feature will survive many unintended mistakes by future contributions

mofosyne (commented):
tests added

PR #255, opened 2023-08-24
add stopgap to_dimensional()

mofosyne:
This function should really be to_dimensional(), but to_dimensional() doesn't exist until inkscape v1.2 and ubuntu package manager only support up to that point so... best to add this stopgap function for now


mofosyne:
closed because this approach doesn't work for regmark rendering as viewport_to_unit() has no simple equivalent that can be added as a inline dummy in inkscape v1.1

Issue #259, opened 2023-08-27
ImportWarning: DynamicImporter.exec_module() not found and mathplotlib depreciation warning


mofosyne:
inkscape console:

```
<frozen importlib._bootstrap>:671: ImportWarning: DynamicImporter.exec_module() not found; falling back to load_module()
<frozen importlib._bootstrap>:671: ImportWarning: DynamicImporter.exec_module() not found; falling back to load_module()
/home/mofosyne/.local/lib/python3.10/site-packages/matplotlib/backends/_backend_gtk.py:146: DeprecationWarning: Gtk.Window.set_wmclass is deprecated
  self.window.set_wmclass("matplotlib", "Matplotlib")
```

logfile: 

```
Inkscape 1.3 (1:1.3+202307231459+0e150ed6c4)
Inkex: 1.3.0
Inkscape-Silhouette: 1.28
Path: /usr/share/inkscape/extensions/sendto_silhouette.py
Python: /usr/bin/python3
Version: 3.10.12 (main, Jun 11 2023, 05:26:28) [GCC 11.4.0]
Platform: linux
Arguments: sendto_silhouette.py --id=path1 --active-tab=logdump --x_off=0.00 --y_off=0.00 --cuttingmat=cameo_12x12 --toolholder=1 --tool=default --media=300 --speed=0 --pressure=0 --depth=-1 --preview=true --dashes=false --autocrop=false --bbox-only=false --multipass=1 --reversetoggle=false --endposition=start --end_offset=0.0 --regmark=true --regsearch=true --regoriginx=0.0 --regoriginy=0.0 --regwidth=0.0 --reglength=0.0 --wait_done=false --sharpencorners=false --sharpencorners_start=0.1 --sharpencorners_end=0.1 --overcut=0.5 --strategy=zorder --orient_paths=natural --fuse_paths=true --sw_clipping=true --logfile=/tmp/silhouette.log --log_paths=false --append_logs=false --inc_queries=false --dry_run=false --force_hardware=DETECT --bladediameter=0.9 /tmp/ink_ext_XXXXXX.svgJW4CA2
Using Registration Mark:: mark distance from document: Left=0mm, Top=0mm; mark to mark distance: X=210.0mm, Y=297.0mm;
Preview aborted.
 done. 0 min 1 sec
```

OS Spec (via `neofetch --off`):

```
OS: Linux Mint 21.2 x86_64 
Kernel: 5.15.0-79-generic 
[...]

PR #262, opened 2023-09-02
Require gtk4

mofosyne:
This tackles https://github.com/fablabnbg/inkscape-silhouette/issues/259

Arguably we could just wait for this to be resolved, but up to you.


mofosyne:
(Not sure why `sendto_silhouette.inx` is in this commit, but it's looks already fixed in main)

mofosyne:
Huh? No gi module?

PR #308, opened 2024-09-22
fix: remove call to missing function with unblock_inkscape

### LINKED ISSUE #213: Multi Action Does Nothing
Version: Inkscape: 1.2, extension: latest commit (0c8942bffdd2d30136bbad41fdb4e50553ce4dec)
OS: Ubuntu 22.04
Hardware: Portrait 3

When I press the Execute button, nothing happens. The machine does not do anything, and the window does not change in any way. The cancel button does not do anything either. Manually cutting twice works, so the issue is not with the machine. I'm not sure how to access any possible error logs since Inkscape doesn't remain attached to the terminal.

TapuCosmo:
Fixes #213. Multi Action now works without blocking Inkscape.

A call to a nonexistent function `show_log_as_dialog` was previously being made if `unblock_inkscape` was true, which caused Multi Action to fail when not blocking Inkscape. Removing this call allows Multi Action to work properly without blocking Inkscape.

### DIFF
--- a/silhouette/MultiFrame.py
+++ b/silhouette/MultiFrame.py
@@ -296,8 +296,6 @@ class MultiFrame(wx.Frame):
         # end wxGlade
 
     def wrapup(self):
-        if self.options.unblock_inkscape:
-            show_log_as_dialog(self)
         self.Close()
 
     def close(self, event):


lex-neufeld:
@TapuCosmo What steps or test can reproduce the error? for #308

this function is in silhouette_multi.py on line 41 https://github.com/fablabnbg/inkscape-silhouette/blob/1e8261eeac70e753b5df2415cc0e896fd67ae994/silhouette_multi.py#L41 

It seems to be desired that a dialogue box of "Silhouette Multi Log" be shown anytime a background process for the Silhouette finishes, or anytime you try to close inkscape if any background processes are running. Something like that. I don't really understand why or exactly when it should display, but it seems clear that we should make the function call work (as it probably did prior to june 2022) rather than remove it. I will open a new issue in a moment and tag you in it.

easyname889:
Still broken at HEAD 4734b8d: `MultiFrame.wrapup()` calls `show_log_as_dialog`, which is undefined in that module (it lives in `silhouette_multi.py`), so every job from Silhouette Multiple Actions in default unblock mode ends in NameError after cutting finishes.

Reproduced on macOS (Darwin 25.5) + Inkscape 1.4.4 + Silhouette Cameo 5 Alpha Plus over BLE, while running 330 gsm boxboard production sheets through Multi Action.

This PR's two-line removal fixes it — patched locally and verified in production since 2026-08-22. @fablabnbg the target lines are still verbatim at HEAD, so this applies cleanly; happy to rebase if that helps get it merged.

Issue #279, opened 2023-11-29
ModuleNotFoundError: No module named 'gi'


brandonros:
Opening an issue for myself to work through to try and help get a clean run on this with Mac.

```
Traceback (most recent call last):
  File "/Applications/Inkscape.app/Contents/Resources/share/inkscape/extensions/inkex/gui/__init__.py", line 39, in <module>
    import gi
ModuleNotFoundError: No module named 'gi'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Applications/Inkscape.app/Contents/Resources/share/inkscape/extensions/inkman/inkman/manage_extensions.py", line 29, in <module>
    from inkex import gui
  File "/Applications/Inkscape.app/Contents/Resources/share/inkscape/extensions/inkex/gui/__init__.py", line 50, in <module>
    raise DependencyError(
inkex.utils.DependencyError: You are missing the required libraries for Gtk. Please report this problem to the Inkscape developers.
```

Happens when going to Extensions -> Manage Extensions

Also worth nothing Extensions -> Render -> Silhouette is not there

t0b3:
> Also worth nothing Extensions -> Render -> Silhouette is not there

is the INX file in the wrong folder maybe? you can double check what is the folder given in the inkscape > preferences > system > inkscape extensions and verify the installation of `sendto_silhouette.inx` went to the correct place... HTH

brandonros:
Inkscape extensions: `/Applications/Inkscape.app/Contents/Resources/share/inkscape/extensions`

```
INFO: installing extension
INFO: directory already exists '/Users/brandon/Library/Application Support/org.inkscape.Inkscape/config/inkscape/extensions/silhouette'. Removing it and recreating it.
```

It's getting installed to `User extensions` folder instead, do you think that's an error or ok @t0b3 ?

![image](https://github.com/fablabnbg/inkscape-silhouette/assets/8949910/929d8eb9-edc2-4b1e-ab06-a8b4fc3e2684)

brandonros:
https://inkscape.org/forums/questions/manage-extensions-not-working-version-11-no-module-named-gi/ it could be an Inkscape extension manager issue and not related to this extension per se

t0b3:
> It's getting installed to `User extensions` folder instead, do you think that's an error or ok @t0b3 ?

both should work fine, although could be interesting for debugging purpose to try it the other way round to see if the observed error persists...

brandonros:
```
[...]

PR #330, opened 2025-04-03
Added missing regmarks scripts copy to osx bash script

grepson:
Basically in macos there is no copy of regmarks scripts while installing the package.
Both files `"render_silhouette_regmarks.inx"` and `"render_silhouette_regmarks.py"` needs to be copied on install in order to fix missing render tab: https://github.com/fablabnbg/inkscape-silhouette/issues/279#issue-2016831816


t0b3:
@grepson THX for your contribution :smile:

grepson:
@t0b3 You're welcome <3

PR #335, opened 2025-06-30
Add support for Portrait4 and 8.5 x 12" cutting mat

cyr123:
* Portrait 4 plotter added to udev rules and main source code.
* Portrait 4 cutting mat (type 11) added to source code and as a template SVG.

Note that it was already possible to use this machine, but with some limitations (most notably the autoblade wasn't being set). The machine has been added as basically a Portrait3 but with slightly wider cutting area.

From my testing and comparison with Silhouette Studio from a small USB capture I took I believe this is all correct.


t0b3:
@cyr123 THX for your valuable contribution 🥇

PR #352, opened 2025-11-13
Fix Externally Managed Environment - Update OS X Install Script - Uses Inkscape's Python

sam-itt:
This pull request fixes the homebrew error "Externally Managed Environment" on recent OS X versions. Rather than using homebrew's python and having to re-install/duplicate all packages provided by Inkscape, it creates a venv that is a strict superset of the Inkscape bundled python and installs additional dependencies safely onto that.

Additionnaly it updates the readme and adds safeguards on the scripts to avoid direct calling of the python install script.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -1,6 +1,7 @@
 name: Run Python tests
 
 on:
+  workflow_dispatch:
   push:
     branches:
       - main
@@ -61,7 +62,7 @@ jobs:
       if: ${{ matrix.os == 'macos-latest' }}
       run: |
         brew install --cask inkscape
-        pip install libusb1
+        brew install libusb
     - name: 🗖 install Inkscape (fixed)
       if: ${{ matrix.os == 'windows-latest' && startsWith(matrix.inkscape-version, '1.') }}
       run: |
@@ -73,13 +74,22 @@ jobs:
     - name: inkscape version
       run: |
         inkscape --version || echo "inkscape not found"
-    - name: install requirements
+    - name: install requirements (macos)
+      if: ${{ matrix.os == 'macos-latest' }}
+      run: |
+        ./install_osx.sh
+        . "$HOME/.local/share/venvs/inkscape/bin/activate"
+        pip install pytest
+    - name: install requirements (non-macos)
+      if: ${{ matrix.os != 'macos-latest' }}
       run: |
         python -m pip install --break-system-packages --upgrade pip
         pip install -r requirements.txt
         pip install pytest
     - name: test
+      shell: bash
       env:
         PYTHONWARNINGS: default
       run: |
-        pytest -s -vv test
+       [[ -e "$HOME/.local/share/venvs/inkscape/bin/activate" ]] && source "$HOME/.local/share/venvs/inkscape/bin/activate"
+       pytest -s -vv test
--- a/README.md
+++ b/README.md
@@ -145,14 +145,15 @@ Finally, load the file with:
 <summary>Click to get steps</summary>
 
 * Install prerequisites:
-  * install homebrew http://brew.sh/
+  * Install homebrew http://brew.sh/
   * `brew install libusb`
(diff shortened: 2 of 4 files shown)


t0b3:
@sam-itt great idea :tada: 

can you ensure this PR
- drops all the unneeded instructions i.e. `brew install python` as it aims to use the one package with inkscape IIUC
  - what about libusb1?
  - nb: can you add in the instructions how to install inkscape? using brew or how?
  - and what about brew itself after all? is it still required?
- update the github action to enable testing of that very proposed setting

this will help to ensure everyone gets a minimal common understanding and have that tested using CI :smile:

sam-itt:
This PR now:

- Updates the README to:
    - Remove unused install instructions, i.e. install python with homebrew
    - Add instructions on how to install Inkscape
- Updates the install script (py) to pull all the required deps, including wxpython and mathplotlib. Numpy removed as a hard dep as it is (also?) a child dep.
- Update the github action to actually test the new setup.

libusb(python) is still pulled from the install script. I've noted that the readme called for installing libusb(binary) with brew while the actions had a pip install. The pip install is/was already done by the original install script. I've followed the readme at first so I installed the binary lib with homebrew and I would say it is a requirement. I did not try to remove it as it could break my system.

I think brew is needed to provide libusb. I use it on my system and cannot try to remove it, for obvious reasons :).

I've tried the setup I provide with a Cameo 4 and it seems to work. Didn't test all and every feature though as I've only used it to cut very simple path at the moment (round gaskets).

.github/workflows/run-tests.yml
```diff
@@ -73,12 +74,27 @@ jobs:
...
-    - name: test
+    - name: test (macos)
+      if: ${{ matrix.os == 'macos-latest' }}
+      env:
+        PYTHONWARNINGS: default
+      run: |
+        . "$HOME/.local/share/venvs/inkscape/bin/activate"
+        pytest -s -vv test
+    - name: test (non-macos)
+      if: ${{ matrix.os != 'macos-latest' }}
```
t0b3: @sam-itt nit: can we have a uniform test step and avoid having distinct steps for macos and non-macos? (i.e. foresee other platforms might also go the road of venv...)
i.e. some try: activate ... finally: pytest logic

.github/workflows/run-tests.yml
```diff
@@ -86,16 +86,12 @@ jobs:
...
-        pytest -s -vv test
-    - name: test (non-macos)
-      if: ${{ matrix.os != 'macos-latest' }}
+    - name: test
       env:
         PYTHONWARNINGS: default
       run: |
-        pytest -s -vv test
+       set -e
+       if [[ "$RUNNER_OS" == "macOS" ]]; then
```
t0b3: @sam-itt could you detect if this file exists - and if so source it? ...so it will work platform independently

.github/workflows/run-tests.yml
```diff
@@ -91,7 +91,6 @@ jobs:
         PYTHONWARNINGS: default
       run: |
```
t0b3, line 93: wondering if it would help to add configuration `shell: bash`
and then to use `[[ -e ~/.local/share/venvs/inkscape/bin/activate ]] && \`

sam-itt:
The action code now uses python to detect the platform and do the activation only on OS X and keep the previous behavior without duplicating the "test" step.

I could not get to something acceptable using shell-based switching. I'm fairly new to github actions so maybe I'm missing something.

t0b3:
I'm wondering if it would work with setting shell to bash and sourcing if file exists
```yaml
    - name: test
      shell: bash
      env:
        PYTHONWARNINGS: default
      run: |
        [[ -e "$HOME/.local/share/venvs/inkscape/bin/activate" ]] && source "$HOME/.local/share/venvs/inkscape/bin/activate"
        pytest -s -vv test

```

sam-itt:
That seems to work

t0b3:
THX for your contribution :tada:

PR #369, opened 2026-05-18
add hardware-native Bezier BZ1 manual

DSGM-GMSD:
## Summary

Adds a new documentation page describing the hardware-native cubic Bezier spline command in the Graphtec GP-GL protocol.

- Documents the verified `BZ1` command syntax and coordinate/parameter mapping for hardware-rendered cubic Bezier curves.
- Explains why offloading curve interpolation to the firmware interpolator reduces USB payload size and avoids serial-buffer stuttering.
- Includes a Python example that generates a native `BZ1` GP-GL command string.

This manual is scoped to the `BZ1` Bezier command only; physical/coordinate clarifications for other commands are intentionally out of scope for this PR.

## Test Plan

- [x] Document renders correctly in Markdown.
- [x] The Python `BZ1` generation example runs and produces the expected output string.

## Type of change

- [x] Documentation

🤖 Generated with [Hermes Agent](https://hermes-agent.nousresearch.com)


DSGM-GMSD:
The commands were physically tested on a Cameo Pro Mark II

t0b3:
> The commands were physically tested on a Cameo Pro Mark II

@DSGM-GMSD can you add just this remark on the very device you tested in the markdown? :+1:

PR #370, opened 2026-06-14
feat: Add support for Silhouette Cameo 5 Alpha Plus

ndahlquist:
## Summary

- Adds the Silhouette Cameo 5 Alpha Plus (USB VID `0x3844`, PID `0x0002`) as a recognized device
- Device uses the same vendor ID as the Cameo 5 Alpha (`0x3844`) with the next sequential product ID (`0x0002`), confirmed via USB enumeration against a live unit (firmware `V1.03`)
- Hardware spec inherits from the Cameo 5 Alpha with width bumped to 372mm to match the \"Plus\" form factor (same as Cameo 5 Plus)

## Changes

- **`silhouette/Graphtec.py`**: New `PRODUCT_ID_SILHOUETTE_CAMEO5ALPHA_PLUS = 0x0002` constant; added to `PRODUCT_LINE_CAMEO4` for Cameo 4-gen command compatibility; hardware entry with `width_mm=372`, `max_pressure=40`, `quadregmarks=True`
- **`silhouette-udev.rules`**: Added `0x3844/0x0002` entry for Linux USB permission access
- **`sendto_silhouette.inx`**: Added Cameo 5 Alpha Plus (and the previously missing Cameo 5 Alpha) to the Inkscape device dropdown
- **`README.md`**: Added Cameo 5 Alpha Plus to the supported devices list

## Test plan

- [x] Device detected as `Silhouette_Cameo5_Alpha_Plus` via `SilhouetteCameo()` on macOS
- [x] `status()` returns `unloaded`, `get_version()` returns `CAMEO 5 ALPHA PLUS V1.03`
- [x] Verify `width_mm=372` is accurate by running a cut that reaches the right edge
- [x] Verify `margin_left_mm=-6.0` produces correct registration on physical media

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ndahlquist:
Tested via Python script on my Silhouette Cameo 5 Alpha Plus. 

Note: I have not tested this via Inkscape (I intend to only use this via Python), but presumably it should work.

PR #375, opened 2026-07-04
feat: add bluetooth support

### LINKED ISSUE #88: Please add Bluetooth support
It would be nice, if Bluetooth interface of newer plotters could be used.
My Silhouette Portrait 2 supports Bluetooth, and I was able to pair the device using the Linux Bluetooth desktop utilities. With pairing the Bluetooth LED on the plotter turns from blue to violett (a PIN was not needed).
The plugin do not support that interface yet, but a support would be nice.

misspokeycat:
## What Changed
- Add built-in Bluetooth support.
- Bluetooth is added via Bluetooth classic - BLE exists for the cutters, but 1. doesn't come out of the box for Python under `socket` and 2. isn't nearly as stable of a connection for sending large jobs.
- Add new Bluetooth tab to the extension for configuring Bluetooth.
  - If `bluetooth_addr` is specified, cutter sends over Bluetooth.
  - Add small "scan" path/option (`bluetooth_scan`) for making it easier to determine your cutter's BT MAC.
- Refactor USB communication into a new `Transport` class alongside Bluetooth comms.
- Add device detection for Bluetooth
  - Based on observed behavior in Silhouette Studio, Studio determines connected cutter type when connected to BT by name, so we do the same here
 - Add new translation strings to `.po` files.

## How It Was Tested
- Tested with my Cameo Alpha 5a, correctly identified via version commands, cutting works as expected.
- Added new test suite for bluetooth
- All existing tests pass

## Known Limitations
- Windows and Linux both allow `socket.AF_BLUETOOTH`, but MacOS _does not_.  It might be possible to get this working via another path on Mac, since I _think_ MacOS opens a socket device when you pair/connect with a cutter, but I have no way of testing this.

Closes #88

### DIFF
--- a/README.md
+++ b/README.md
@@ -236,6 +236,12 @@ Refer to the [userguide instructions](./USERGUIDE.md) for further details.
 
 Run `sendto_silhouette.py --help` for information on CLI usage.
 
+### Bluetooth
+
+Bluetooth-capable cutters (Cameo 3 and newer, Portrait 2 and newer, Curio 2, and similar) can be driven wirelessly instead of over USB. Pair the cutter with your operating system first, then use the **Bluetooth** tab in the GUI (or `--bluetooth_scan` / `--bluetooth_addr` on the CLI). See [Connecting over Bluetooth](./USERGUIDE.md#connecting-over-bluetooth) for step-by-step instructions.
+
+Bluetooth is currently supported on **Linux and Windows only**; macOS lacks Python RFCOMM socket support, so USB must be used there for now (contributions to add macOS Bluetooth are welcome).
+
 ---
 
 ## Templates
@@ -288,6 +294,9 @@ This fails on win32/64 with 'module has no attribute 'version info' which then c
 * reverse toggle options, to cut the opposite direction. This might also be
   helpful with mat-free cutting via multipass.
 * honors hidden layers.
+* Bluetooth support. Cutters with Bluetooth can be driven wirelessly instead of
+  over USB, after pairing with the operating system. See
+  [Connecting over Bluetooth](./USERGUIDE.md#connecting-over-bluetooth).
 
 ## Misfeatures of InkCut that we do not 'feature'
 
--- a/USERGUIDE.md
+++ b/USERGUIDE.md
@@ -83,6 +83,46 @@ For enhanced precision, you may have to set an offset on **X-Offset** and/or **Y
 
 ---
 
+## Connecting over Bluetooth
+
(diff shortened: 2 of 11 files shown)


t0b3:
@misspokeycat great contribution :smile: 
can you add actual translations to your PR?

misspokeycat:
> @misspokeycat great contribution 😄 can you add actual translations to your PR?

@t0b3 Only if machine translations are acceptable in this project - if so, I can add them with a note that they need review.  I only natively speak English, if native speakers of the other languages want to check/provide better ones, that would be welcomed :)

pkoryzna:
wow nice work! 

regarding BLE i have a fork that mostly works, it is using an external python lib in a subprocess and sends it data over a tcp socket. kinda hacky, but it works on macOS with my Cameo 4 - https://github.com/pkoryzna/inkscape-silhouette-bt

funny part is i got stuck on the gui/scanning part a year ago, and literally got a gui for device working today, so maybe you can use some parts of that? https://github.com/pkoryzna/inkscape-silhouette-bt/blob/update_ble/silhouette/BluetoothScan.py

misspokeycat:
> regarding BLE i have a fork that mostly works, it is using an external python lib in a subprocess and sends it data over a tcp socket. kinda hacky, but it works on macOS with my Cameo 4 - https://github.com/pkoryzna/inkscape-silhouette-bt

I had seen the work done in the issue thread for this, though opted purposely not to use `ble-serial` since it would add another dependency to the project (alongside the stability difference of BLE vs BT classic).  That being said, it might be unavoidable for MacOS support to pull in another library for this, since we can't use `socket AF_BLUETOOTH` on MacOS.  I'd half imagine a clean path to do that is add another transport for a generic serial port, and then we could leverage whatever MacOS opens, or just use an external program/library like `ble_serial` for it.  We could also just have it where Mac users need an additional dependency for Bluetooth (likely just drop in Pybluez since that's useful for platform-agnostic discovery), optional on other platforms.

> funny part is i got stuck on the gui/scanning part a year ago, and literally got a gui for device working today, so maybe you can use some parts of that? https://github.com/pkoryzna/inkscape-silhouette-bt/blob/update_ble/silhouette/BluetoothScan.py

There's a GUI integrated into this PR for finding your cutter's BT mac address, though it's a bit bare-bones.  Sadly, `.inx` extensions can't dynamically update their contents, I'd have liked to just scan for a dropdown if I could.

t0b3:
@misspokeycat great contribution :tada: big thanks

PR #382, opened 2026-08-06
Add macOS Bluetooth Low Energy transport

kvasdopil:
## Summary

Add Bluetooth Low Energy as a third connection transport alongside USB and Bluetooth Classic/RFCOMM, with initial macOS support through `bleak` and CoreBluetooth.

## What changed

- Add a synchronous BLE transport backed by a dedicated asyncio worker thread.
- Prefer peripherals advertising the known vendor service and validate the connected GATT service before initialization.
- Scan once for the selected device, prioritize an explicit identifier or name over the generic filter, and stop immediately when an exact identifier is advertised.
- Pass the discovered `BLEDevice` to `BleakClient`, avoiding Bleak's implicit second scan for string identifiers.
- Implement GATT subscription and initialization, acknowledged 20-byte writes, buffered reads, timeout handling, and deterministic shutdown.
- Keep control-characteristic movement notifications separate from command responses received on the read characteristic.
- Retry the cutter's transient ATT `0x0e` / “Unlikely Error” backpressure response with bounded exponential delay.
- Wire BLE selection into the Graphtec driver while preserving the existing USB and RFCOMM paths.
- Add Inkscape UI and CLI options for connection type, advertised name, local identifier, and unfiltered device scanning.
- Default the advertised-name selector to `CAMEO`.
- Check cutter status before setup or plot commands and exit early with a clear message when no media is loaded or status cannot be confirmed.
- Prevent macOS system sleep for the complete physical cut using a PID-bound `caffeinate` child, with cleanup on normal completion, early returns, exceptions, and interruption.
- Remove the experimental `-6 mm` horizontal margin from the Cameo 5 Alpha and Cameo 5 Alpha Plus profiles.
- Install `bleak` and its required extension dependencies through the macOS installer.
- Add fake-stack transport and Graphtec integration tests.
- Document separate Bluetooth Classic and BLE workflows, macOS permissions and installation, BLE device selection, and the no-media early exit in the README and User Guide.

## Why

Recent Silhouette cutters expose the Graphtec command stream over a vendor BLE GATT service. On macOS, CoreBluetooth exposes a computer-local identifier rather than a portable MAC address, so the extension needs name-based discovery and a BLE-specific transport instead of reusing the RFCOMM address workflow.

Live Cameo 5 Alpha diagnostics also exposed two correctness issues:

1. The control characteristic emits short movement events such as `b'1\x03'` and `b'0\x03'`, while query and registration results arrive on the read characteristic. Merging both streams allowed the initial “moving” event to be consumed as a registration result, causing an immediate false failure.
2. The Cameo 5 Alpha profiles contained an experimental `-6 mm` left margin. The normal Inkscape plotting path applied it to cut geometry, shifting cuts horizontally relative to the document and registration marks.

The initial identifier optimization also needed correction: constructing `BleakClient` with a string identifier does not bypass discovery—Bleak performs an implicit scan. A direct-connect attempt followed by fallback could therefore scan twice and could leave a partially connected client behind. In addition, a generic `CAMEO` match could end discovery before the explicitly selected identifier appeared. The revised flow performs one explicit, early-stopping scan and always connects using the discovered device object.

## User impact

macOS users can select Bluetooth Low Energy in the Send to Silhouette extension, grant the standard macOS Bluetooth permission, scan nearby devices, and target a cutter by its advertised name or local identifier. When a saved identifier is used, discovery stops as soon as that exact device appears. Existing USB and Bluetooth Classic behavior remains available.

Jobs stop before device setup or plotting if the cutter reports that no media is loaded, avoiding a partially started job. BLE registration waits for the actual command response instead of failing on a control-channel movement event, and Cameo 5 Alpha cuts no longer receive the experimental 6 mm horizontal shift.

During real cutting jobs, macOS is kept awake until the cutter reports that it is no longer moving. The extension terminates only the `caffeinate` process it started; binding that child to the extension PID also releases the assertion if the extension is forcibly terminated.

## Validation

- 107 tests pass; 1 hardware-dependent test is skipped.
- Regression coverage verifies that an unrelated Cameo advertisement cannot end discovery before a delayed explicitly selected device appears.
- Sleep-prevention coverage verifies normal cleanup, dry-run behavior, launch failure, forced cleanup after a timeout, and cleanup-error suppression.
- Ruff lint and formatting checks pass for the touched BLE transport and test modules.
- Python compilation and Inkscape `.inx` XML validation pass.
- Inkscape 1.4.4 on macOS successfully requested Bluetooth permission and displayed an unfiltered list of nearby BLE devices.
- The full flow was verified end-to-end over BLE on a Silhouette Cameo 5A with the pen attachment, from device discovery and connection through status checking, setup, and drawing.
- Live BLE tracing on a Cameo 5 Alpha confirmed that movement events arrive on the control characteristic and command/registration responses arrive on the read characteristic.
- Physical diagnostic cuts confirmed that the profile's `-6 mm` margin was applied to ordinary Inkscape cut coordinates and was not needed for the tested Print & Cut setup.

## Protocol attribution

Characteristic roles and the three-characteristic initialization handshake were cross-checked against [`bob-takuya/cameo-cut`](https://github.com/bob-takuya/cameo-cut) (MIT).

## Still to verify

- Successful four-point optical registration with the current generated marks; after response routing was corrected, the cutter waited for and returned a genuine optical-search result rather than failing immediately on a movement event.
- Broader Cameo model and attachment coverage, plus retry behavior for transient scan/connect failures.


deveth0:
Tested with a cameo 4,  works like a charm!

kvasdopil:
Tested Print-and-cut with regmarks on Cameo 5, found a couple of bugs, should be fixed now.

deveth0:
@kvasdopil I added a PR to your branch which improves the performance when connecting to the device.

deveth0:
I found another issue in long running jobs.
MacOS kills the connection after a certain idle time of the system. 

I think we should probably wrap the process into caffeinate to ensure, it finishes successfully.

kvasdopil:
> I found another issue in long running jobs. MacOS kills the connection after a certain idle time of the system.
> 
> I think we should probably wrap the process into caffeinate to ensure, it finishes successfully.

Or maybe smth like periodic pings over the control channel could solve the issue. How do i reproduce?

deveth0:
For me it was enough to start a job that took around 5 minutes and have my Mac idle (but connected to power).
It then failed with Bluetooth connection failed (or so).

kvasdopil:
> For me it was enough to start a job that took around 5 minutes and have my Mac idle (but connected to power). It then failed with Bluetooth connection failed (or so).

@deveth0 added sleep prevention. Can you try new version?

deveth0:
Hey, I already did this in an open pr 😁

kvasdopil:
@deveth0 yup, i also added a few more guars to make sure we're not keeping caffeinate if smth goes sideways. 

@t0b3 wdyt? can we merge that in?

deveth0:
Will install and test later

t0b3:
@deveth0 your test is very welcome :+1:

deveth0:
looks good to me, also gave it a shot and works like a charm!

t0b3:
@misspokeycat would this PRs BLE implementation work for your environment as well? - would be great to have your feedback and remarks or PR on this :smile:

